PR #4, opened 2011-11-03
Fix indentations in celery plugin_health_check.py

PamelaM:
I fixed that, I also added /build, *~ and *# to .gitignore


KristianOellegaard:
Thanks a lot for the fix, happy to see that other people are using it as well :-)

PR #5, opened 2013-03-28
Added health_check.utils.healthcheck

ojii:
Added a decorator that makes it a lot easier to write checks. Inside the
decorator you can raise health_check.utils.HealthCheckFailed or
health_check.utils.HealthCheckUnavailable instead of returning
status values.
The decorator optionally accepts a string with a descriptive name of the
check.

### DIFF
--- a/health_check/utils.py
+++ b/health_check/utils.py
@@ -6,21 +6,8 @@ from health_check.plugins import plugin_dir
 
 class BaseHealthCheck(BaseHealthCheckBackend):
     def check_status(self):
-        try:
-            self._wrapped()
-        except HealthCheckFailed:
-            return HealthCheckStatusType.unexpected_result
-        except HealthCheckUnavailable:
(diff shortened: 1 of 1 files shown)


health_check/utils.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+from django.db.models.options import get_verbose_name
+from health_check.backends.base import HealthCheckStatusType, BaseHealthCheckBackend
+from health_check.plugins import plugin_dir
+
+
+class BaseHealthCheck(BaseHealthCheckBackend):
+    def check_status(self):
+        self._wrapped()
+        return HealthCheckStatusType.working
```
KristianOellegaard: Still not a big fan of this. Btw its fine to return `True` instead of `HealthCheckStatusType.working` (True is encouraged).

PR #7, opened 2013-11-07
Django 1.6 compatibility

chripede:
django.conf.urls.defaults was removed with django 1.6

### DIFF
--- a/health_check/urls.py
+++ b/health_check/urls.py
@@ -1,4 +1,4 @@
-from django.conf.urls.defaults import patterns, include, url
+from django.conf.urls import patterns, include, url
 
 import health_check
 health_check.autodiscover()


KristianOellegaard:
I'm so sorry! Completely missed this!

Will this work with releases before 1.6 as well or should I make a try/except to make sure it works on both?

chripede:
django.conf.urls.defaults was deprecated back in 1.4 so this patch should work from there. You list django 1.2+ as a requirement though, so either try/except or update requirements :)

KristianOellegaard:
I think it's fine to go for 1.4+, I'll update requirements

KristianOellegaard:
https://pypi.python.org/pypi/django-health-check/0.3

chripede:
Thank you

README still says 1.2+ btw.

PR #8, opened 2013-12-16
Make django-health-check 1.6 compatible

albandum:
Django 1.6 has some non-backwards compatible moves, you might want to keep 2 different branches.


KristianOellegaard:
Hi there,

Master branch should now work with 1.4+ (including 1.6)

Sorry about the late answer

Kristian

PR #9, opened 2014-02-07
Plugin health checks as json rather than HTML

philroche:
Added a simple view to return the plugin health checks as json rather than HTML


helgi:
This seems pretty useful to me. Could it be rebased and reviewed? :)

health_check/backends/base.py
```diff
@@ -29,15 +29,23 @@ class ServiceReturnedUnexpectedResult(HealthCheckException):
 
 class BaseHealthCheckBackend(object):
 
+
+    def description(self):
+        return None
```
ghost, line 34: I'd propose to default to `self.__class__.__doc__` That way if you have a docstring on the HealthCheck class it will be used.

Also I'm a bit confused between BaseHealthCheck and BaseHealthCheckBackend. Why BaseHealthCheckBackend does not extend BaseHealthCheck ?

health_check/management/commands/healthcheck.py
```diff
@@ -0,0 +1,16 @@
+__author__ = 'philroche'
+import json
+import sys
+from django.core.management.base import BaseCommand, CommandError
+from health_check.utils import get_plugins
+
+class Command(BaseCommand):
```
ghost, line 7: This command does not have anything to do with JSON responses.

health_check/management/commands/healthcheck.py
```diff
@@ -0,0 +1,16 @@
+__author__ = 'philroche'
```
ghost, line 1: **author** is extraneous and the author seems to be following PEP8 recommendation to group imports into 3 groups. http://legacy.python.org/dev/peps/pep-0008/#imports

ghost:
Disclaimer: I'm not the maintainer but I took the liberty to add some notes.

I know this is sort of a old PR. Though this is a needed feature for this library to really be useful. Thanks for taking a dive at it. I'm considering using one of the forks that adds this feature until ones gets merged. I think it would be wrong to reimplement it yet another time.

This PR can be broken down into 3. Fixing compatibility issues, creating a command that runs checks from the command line, and adding different content types (json, yaml, and plain text).

For the rest I'll let @KristianOellegaard weighing in.

ghost:
If no one volunteers I could take some time and cherry pick a few commits from this PR to create smaller easier to merge PR. As well as rebasing.

codingjoe:
@jackdbernier be my guest, feel free to do that.

PR #16, opened 2014-12-02
Fix bad return of ServiceUnavailable exception

failedguidedog:
The ServiceUnavailable exception in StorageHealthCheck was not raised correctly which resulted in `KeyError: ServiceUnavailable('Unknown exception',)` exception when DefaultFileStorageHealthCheck failed.


KristianOellegaard:
Thanks!

PR #21, opened 2015-01-28
Improve README

jairhenrique:


codingjoe:
Sorry, this seems to be very old. Most of the badges no longer work. I merged what I could and change the badges

Thanks!

PR #22, opened 2015-01-28
Add more classifiers on setup.py

jairhenrique:


codingjoe:
Thanks, I merged the classifier part

PR #34, opened 2015-12-08
Keep __context__ for custom exeptions.

codingjoe:
This is py3 only.

Since python there is an option to keep the `__context__`
when reraising exceptions. This may only change the exception
type but not the context when and where it was raised.


codingjoe:
Closed as project seems unmaintained.

PR #40, opened 2016-04-13
Major Refactoring

codingjoe:
While we upgraded our stack to py35 and dj19 we did some major changes to django-health-check.

Most notably made it work similar to Django's form validation.

`check_status` doesn't return anything, but raises custom `Exceptions` now.

I have a couple of py3 only features in there like `raise from` but I can get rid of that if you'd like me to.

Further more, if you don't feel like maintaining this any longer, let me know. As we're using it at my company, we can maintain it for the future.

Cheers,
Joe

### DIFF
--- a/health_check/backends/base.py
+++ b/health_check/backends/base.py
@@ -1,4 +1,10 @@
+import logging
+
+from django.utils.encoding import force_text
+from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
+
+logger = logging.getLogger('health-check')
 
 
 class HealthCheckStatusType(object):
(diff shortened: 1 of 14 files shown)


health_check/backends/base.py
```diff
@@ -1,51 +1,59 @@
...
-HEALTH_CHECK_STATUS_TYPE_TRANSLATOR = {
-    0: _("unavailable"),
-    1: _("working"),
-    2: _("unexpected result"),
-}
-
 
 class HealthCheckException(Exception):
-    pass
+    type = _("unknown error")
```
KristianOellegaard, line 17: type() is a built in, so we should probably go with something different

KristianOellegaard:
Hi there,

Thank you for contributing to django-health-check!

I'm happy to accept this PR if you fix it so it works with python 2 as well :-)

I'm very busy with my startup at the moment, so I'd love to take in another person to help maintaining this project. If you're interested send me an email at kristian@plecto.com and let's take it from there.

Kind regards,
Kristian

adamn:
Any update on this?  Would be nice to get it in.

codingjoe:
@adamn yes, I really want to finish this. It's mostly tests and reviews missing. Currently I didn't find the time to finish it. I'm on vacation right now, but I'll try to look into it asap.

PR #44, opened 2016-07-04
Catch Celery forget() not implemented error

samuelfekete:
When using AMQP as the backend, `result.forget()` raises an error, which causes the test to fail even though Celery is working perfectly fine. In such cases this error should be caught and ignored.


codingjoe:
Hi @samuelfekete, thanks for the contribution.
I know it's much to task, since there are not many tests yet, but can you add a test, that exploits this behavior?
If you are having trouble with that, let me know what setup you need to recreate the bug? Which celery backend and so on.

Thanks,
Joe

codingjoe:
Release in 1.1.2

PR #49, opened 2016-08-25
Added tests for plugins

mspeedy:
Added tests for the database, storage, and cache plugins that mock the backend sources to test healthcheck functionality. All tests written as django TestCases, so the project needed to be a django project to run (as far as I could tell). Added necessary files to make it a runnable django project (manage.py, settings.py, etc.)


health_check/settings.py
```diff
@@ -0,0 +1,129 @@
...
+WSGI_APPLICATION = 'djangomodule.wsgi.application'
+
+
+# Database
+# https://docs.djangoproject.com/en/1.9/ref/settings/#databases
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.sqlite3',
+        'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
```
codingjoe: This should be `:memory:` which will result in the sql db being created in memory only.

health_check/wsgi.py
```diff
@@ -0,0 +1,16 @@
+"""
```
codingjoe: You won't need a WSGI interface. I guess you can just drop the file

codingjoe:
Hi @mspeedy,

thanks for all that. I think it's a really good start. I have a couple of note, to integrate you tests a little smoother.
- py.test is able to run unittests. Thus there is absolutely no problem to mix them.
- As a bonus you don't need to setup a new app just to run the tests. You will need an app tho,
  if you want to test the integration. We have `tests/testapp/` already configured for that.
- Keep in mind the QA tests. You can execute the QA tests by calling `tox -e qa`

Let me know if you need any help. I can also fix everything while merging if that is ok with you.

-Joe

mspeedy:
Hi @codingjoe,
Thanks for your feedback! I'm still relatively new to django/python testing if you couldn't tell, so any help would be greatly appreciated. I've updated the project, removing the extra project files and making the QA tests pass. I'm now running the tests using py.test and all but the autodiscover test are passing, but I am unsure as to why. Let me know what other changes you think should be made.

codingjoe:
@mspeedy ok, I'll try to look into it asap, but it might take a while since I'm about the leave for holiday

codingjoe:
@mspeedy sorry It took me so long to get into this. Can you please rebase, I really want to merge this!

mspeedy:
Hey @codingjoe , This should fix things up. Worth noting though, I moved on to other projects and therefore never resolved the issue with test_autodiscover not passing. I believe the function still works properly, just not the test. You may want to look into it before merging just in case. 

Let me know if there's anything else you need!

aleksihakli:
Hey, seems like [Celery is missing](https://travis-ci.org/KristianOellegaard/django-health-check/jobs/183682633#L404) in tests, should that be added to [requirements-dev.txt](https://github.com/mspeedy/django-health-check/blob/master/requirements-dev.txt) @mspeedy?

codingjoe:
Merged in #69, thanks!

PR #50, opened 2016-08-30
Windows fixes

aleksihakli:
Windows fails on current master branch on Storage health checking.

**Problem 1** is that file names were generated with `datetime.datetime.__str__`, which formats the file names with ISO8601 format which includes colons, `:`, by default, [which are reserved characters in Windows](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx). I have converted the file naming to generate names in alternative [ISO8601 format](https://en.wikipedia.org/wiki/ISO_8601), which should still be equally machine-readable.

**Problem 2** is that file handles that were created for tempfile content validation were not being closed implicitly or explicitly, which is fixed by the included `with open as` method, which guarantees file handles will be closed even when exceptions are raised.

### DIFF
--- a/health_check_storage/base.py
+++ b/health_check_storage/base.py
@@ -38,7 +38,7 @@ class StorageHealthCheck(BaseHealthCheckBackend):
 
     def get_file_name(self):
         return 'health_check_storage_test/test-{}-{}.txt'.format(
-            datetime.datetime.now(),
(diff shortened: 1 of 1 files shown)


codingjoe:
Thanks!
Released in 1.1.4

PR #51, opened 2016-09-13
new plugin for python_rq

Maxtaager:


Maxtaager:
never mind we figure out a better way of doing this.

Issue #62, opened 2016-12-03
Project updates


aleksihakli:
Hi,

I was planning on contributing the following updates to the project:

- Rewrite the bundled storage backend to internally provide and use `check_save` and `check_delete` methods that could be overridden per storage backend instead of overriding the whole `check_status` backend. This would not be a breaking change and would 
- Add a bundled S3 storage backend checker for AWS health checking compatibility, for example `health_check_storage_s3`, which utilizes the S3BotoStorage interface from the django-storages. This is a very widely used package in AWS and Heroku which would be nice to support here as well. It is officially recommended by Heroku;
- Upgrade README from Markdown to reStructuredText for PyPI documentation compatibility (this is the preferred format for Python packages and is supported by GitHub and many other VCS services out-of-the-box.

I thought to ask about the plan before starting the work as to avoid unnecessary changes that could be rejected in PRs.

codingjoe:
@aleksihakli thanks I've seen your PR and left a comment. Please split those things as much as possible. That way we can merge them faster.

aleksihakli:
@codingjoe Hey Joe. I've split the pull requests into 4 smaller PRs, #64, #65, #66 and #67. The logging change, #64, is necessary for cleaner and more concise logging and error handling in the project and I will contribute another set of changes to improve the existing backends if that change gets accepted.

codingjoe:
I haven't looked into the logging PR yet, but I will. You can also check out the [MajorRefactoring PR](https://github.com/KristianOellegaard/django-health-check/pull/40). I did make a lot of changes there, but never found the time to split them into smaller chunks and to merge them.
Maybe you find something interesting in there.

What I currently very much dislike this that some methods return Exceptions.... They should be raised, not returned. Similar to how Validators work in Django.

aleksihakli:
Hey,

I actually saw at least two return statements which are fixed in the S3BotoStorage patch. I'll take a look at the major refactoring patchset and see if it could be rebased into smaller chunks on top of master too.

I'll be contributing some tests later on for a bit more comprehensive coverage.

EDIT: Return behaviour corrected to Raise in here: https://github.com/KristianOellegaard/django-health-check/pull/67/files

codingjoe:
@aleksihakli yeah, the smaller the better and tests if possible. That is the biggest issue really. There are no tests :/

aleksihakli:
All changes merged in the original proposal, so closed.

PR #63, opened 2016-12-04
Refactor project and add S3 storage backend health check

aleksihakli:
Hey,

this PR relates to #62 and attempts to solve the following:

- Add S3 storage backend due to it's widespread use in AWS, Heroku and other hosting providers, and, in addition to that work;
- Convert Markdown README to RST for PyPI page readability; RST is the preferred format for Python packages;
- Deprecate unnecessary models.py files that are not needed for the project;
- Clean up error handling;
- Improve Python 2.6, 2.7 and 3 compatibility in different files by adding missing encoding tags and fixing some typos.

I hope we could merge this to make the S3 storage backend health checking usable in AWS.


codingjoe:
@aleksihakli thanks for the contribution, but this is a bit too much change at once. You do have multiple commits, it would be great if you could split this across multiple pull requests.

aleksihakli:
Split into #64, #65, #66 and #67.

Issue #14, opened 2014-10-21
log exceptions


stefanfoulis:
To make actually debugging an error easier, it would be nice if the error could be logged using standard logging. This would be an alternative approach to #10 .
The "status" could also include the very short version of the exception (or just the Exception class) as information instead of just "unavailable".

aleksihakli:
I think this could be closed as well @codingjoe.

PR #64, opened 2016-12-05
Add logger methods to BaseHealthCheckBackend

aleksihakli:
This patchset introduces logging facilities to base health check backend, which allows subclasses to use uniform logging APIs across the project.

This is base work addressing #14 along with current logging conventions and tries to improve the project code quality.

### DIFF
--- a/health_check/backends/base.py
+++ b/health_check/backends/base.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+from logging import getLogger
+
 from django.utils.translation import ugettext_lazy as _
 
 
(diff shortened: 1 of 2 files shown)


health_check/backends/base.py
```diff
@@ -44,6 +46,23 @@ def status(self):
 
         return self._status
 
+    def get_logger(self, logger):
+        if logger:
+            return logger
+        if hasattr(self, "logger"):
+            return self.logger
+        return getLogger(self.__name__)
```
aleksihakli: We can also use self.__module__ here which should return the child class module from where the logging is done from. This is somewhat equivalent to __name__ used currently in the different subclassed health checking backend modules.

aleksihakli:
Already implemented in a cleaner way in #75.

PR #65, opened 2016-12-05
Improve Python 2.7 compatibility by adding explicit coding to project files

aleksihakli:
This patchset adds unicode coding tags for better Python 2 compatibility.

### DIFF
--- a/health_check/__init__.py
+++ b/health_check/__init__.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Used by setup.py, so minimize top-level imports.
 
 __version_info__ = {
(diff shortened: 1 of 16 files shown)


codingjoe (dismissed):
It's not really needed, since there are no none ASCII characters in those files, but just to be safe it's a good idea.

codingjoe (changes requested):
Please drop the second commit. It has nothing to do with Python 2 support and actually removes support for all Django versions prior to the App loading change in Dajngo 1.7

codingjoe:
@aleksihakli we can talk about removing the old `models.py` files. But that is a breaking chance and would require at least a new major release and some documentation.
I guess we can do this at a later point.

aleksihakli:
Updated PR as per request.

codingjoe:
Released in 1.2.1

PR #67, opened 2016-12-05
Add S3BotoStorage health checker

### LINKED ISSUE #17: DefaultFileStorageHealthCheck fails if backend is Amazon S3
If the backend storage is Amazon S3, the health check fails when it tries to check existence of the deleted test file.

```
            storage.delete(file_name)
            if storage.exists(file_name):
                return ServiceUnavailable("File was not deleted")
```

https://github.com/KristianOellegaard/django-health-check/blob/master/health_check_storage/base.py#L50-L52

The issue is that it take moment for S3 to actually delete the file and therefore the health check fails because the file still exists. It gets to the `exists()` before S3 actually can delete the file because `delete()` on S3 is asynchronous (whereas the `save()` is synchronous because we wait for a file name).

I solved this was sub-classing and not checking if the file was deleted but this isn't an ideal solution. I'd submit a PR if you have a suggestion on how you'd like to fix it. I'd be fine with adding at setting:

```
HEALTH_CHECK_STORAGE_DISABLE_DELETE_EXISTS = True
```

aleksihakli:
Add a backend offering S3BotoStorage health checking support. This patchset depends on #64 and is offered as unstable version at first, hence not documented in README. API does not conflict with the existing APIs nor break backwards compatibility. Fixes #17.

### DIFF
--- a/health_check_storage/base.py
+++ b/health_check_storage/base.py
@@ -45,28 +45,36 @@ class StorageHealthCheck(BaseHealthCheckBackend):
     def get_file_content(self):
         return b'this is the healthtest file content'
 
+    def check_save(self, file_name, file_content):
+        storage = self.get_storage()
+        # save the file
+        file_name = storage.save(
+            file_name, ContentFile(content=file_content)
+        )
+        # read the file and compare
+        if not storage.exists(file_name):
+            logger.exception('File does not exist')
+            raise ServiceUnavailable('File does not exist')
+        with storage.open(file_name) as f:
+            if not f.read() == file_content:
+                logger.exception('File content does not match')
+                raise ServiceUnavailable('File content does not match')
+
+    def check_delete(self, file_name):
+        storage = self.get_storage()
+        # delete the file and make sure it is gone
+        storage.delete(file_name)
+        if storage.exists(file_name):
+            logger.exception('File was not deleted')
+            raise ServiceUnavailable('File was not deleted')
+
     def check_status(self):
         try:
             # write the file to the storage backend
-            storage = self.get_storage()
             file_name = self.get_file_name()
             file_content = self.get_file_content()
-
-            # save the file
-            file_name = storage.save(
-                file_name, ContentFile(content=file_content)
-            )
-            # read the file and compare
-            if not storage.exists(file_name):
-                raise ServiceUnavailable('File does not exist')
-            with storage.open(file_name) as f:
-                if not f.read() == file_content:
-                    return ServiceUnavailable('File content doesn\'t match')
-            # delete the file and make sure it is gone
-            storage.delete(file_name)
-            if storage.exists(file_name):
-                logger.exception('File was not deleted')
-                return ServiceUnavailable('File was not deleted')
+            self.check_save(file_name, file_content)
+            self.check_delete(file_name)
             return True
         except Exception:
             logger.exception('Unknown exception')
--- a/health_check_storage_s3/plugin_health_check.py
+++ b/health_check_storage_s3/plugin_health_check.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+import logging
+
+from health_check.plugins import plugin_dir
+from health_check_storage.base import StorageHealthCheck
+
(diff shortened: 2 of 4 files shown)


aleksihakli:
I removed the dependency for logger updates in #64 for this patchset, allowing it to be merged standalone.

health_check_storage/base.py
```diff
@@ -45,28 +45,36 @@ def get_file_name(self):
...
 
+    def check_save(self, file_name, file_content):
+        storage = self.get_storage()
+        # save the file
+        file_name = storage.save(
+            file_name, ContentFile(content=file_content)
+        )
+        # read the file and compare
+        if not storage.exists(file_name):
+            logger.exception('File does not exist')
```
codingjoe, line 56: I think it might be a good idea to start translating those messages. How about we start using `gettext` as a start?

health_check_storage_s3/plugin_health_check.py
```diff
@@ -0,0 +1,27 @@
...
+
+class S3BotoStorageHealthCheck(StorageHealthCheck):
+    """
+    Tests the status of a `S3BotoStorage` file storage backend.
+
+    S3BotoStorage is included in the `django-storages` package
+    and recommended by for example Amazon and Heroku for Django
+    static and media file storage on cloud platforms.
+
+    django-storages can be found at https://git.io/v1lGx
```
codingjoe: Please use double accents around all technical terms.
eg.
```
``django-strages``
```

health_check_storage_s3/plugin_health_check.py
```diff
@@ -0,0 +1,27 @@
...
+    S3BotoStorage can be found at https://git.io/v1lGF
+    """
+
+    logger = logging.getLogger(__name__)
+    storage = 'storages.backends.s3boto.S3BotoStorage'
+
+    def check_delete(self, file_name):
+        storage = self.get_storage()
+        storage.delete(file_name)
```
codingjoe, line 25: I don't see where you do an assertion here to raise an error. Yes call could raise an error, but you should really check if the file has been deleted.

codingjoe (changes requested):

health_check_storage_s3/plugin_health_check.py
aleksihakli: Please #17 for which this patchset tries to offer a solution for.

---

The S3 storage backend doesn't actually return the deleted item marker or state and the delete operation is asynchronous. The deletion state of the file can not be checked with the current S3BotoStorage storage API. This is the original reason for writing deletion here this way. The old implementation checks if the file has been deleted, but that always raises because the S3 bucket is not in sync yet.


This is due to limitations in the S3BotoStorage interface. I am going to ask for a refactor in that package as well, but that will be a breaking API change, whereas this one should not be.

Possible ways to implement checking if deletion works are:

- Wait for a small while before checking, trying delays between e.g. in 300-3000ms range. This could return false positives due to async calls;
- Implement locks and only allow one check to run here per time. Check if there is more than one file to be deleted. This should work but limits the health checker rate severely.
- Implement the deletion state checker independently of the S3BotoStorage API, fetch delete markers and compare them. This is an extremely volatile approach which is tied to the Boto API version.

So, really, there is no clean way for doing this yet.

health_check_storage/base.py
aleksihakli: Implementing Translation with something that multiple collaborators can neatly use would be smartest to do with Transifex in a separate changeset. Should we add that and translators?

Logged messages shouldn't be translated for log parseability, in my opinion, but UI messages could indeed be.

health_check_storage_s3/plugin_health_check.py
codingjoe: Ok, we shouldn't wait, since this is executed in a request.

health_check_storage/base.py
codingjoe: Yes, we can do this in a separate PR.

codingjoe:
Released in 1.3.0

PR #68, opened 2016-12-14
Added tests for plugins

aleksihakli:
Fixes and help for PR #49 that @mspeedy kindly contributed.

- Add Celery to requirements-dev.txt
- Sort import statements
- Fix counting stock plugin modules in test_autodiscover
- Fixes to Python 2 and 3 compatibility
- Fixes to inherited class initialization semantics (add calls to base classes in `__init__` methods). For some reason QuantifiedCode is still not happy with the Python 2 compatible initialization syntax and suggests just calling `super().__init__`, which we can't do while supporting Python 2 (unless done with [future](http://python-future.org/overview.html) super which would add a dependency).

PR #69, opened 2016-12-15
Added tests for plugins (rebased and updated)

aleksihakli:
Continues where #49 and #68 left off.

The old commit chain had multiple times the necessary commits after rebasing efforts and container some very unclean commits which I've cleaned up. The main beef here is:

- Squash and remove 600kloc library adding commits
- Rebase commits on top of master
- Retains original committer, adds a few necessary migration lines

The patchset tries not to touch other than testing facilities.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -7,3 +7,4 @@ The following people have contributed to django-health-check:
 - Christian Pedersen
 - Johannes Hoppe
 - Aleksi Häkli
+- Mark Speedy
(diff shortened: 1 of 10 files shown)


aleksihakli:
I refactored and revamped the testing patchset and think it's ready for review. There was some work that was overlapping with the current master. 

I wouldn't mind if you reviewed the stuff as author @mspeedy, I had to make some adjustments to your commits and clean them up.

I also changed the tests so that the previously existing tests are not altered and are simple py.test that inherit from objects @codingjoe.

codingjoe (approved):

codingjoe:
@aleksihakli I keep commits to keep the attribution and didn't swash it.

PR #31, opened 2015-07-01
Added support for a JSON health endpoint

clearclaw:
I need machine parseable health reports for our monitoring infrastructure, so I added a JSON endpoint.

PR #72, opened 2016-12-23
Added support for a JSON health endpoint

aleksihakli:
Rebased #31 on top of the current master to get it finally merged.

### DIFF
--- a/health_check/backends/base.py
+++ b/health_check/backends/base.py
@@ -15,6 +15,12 @@ HEALTH_CHECK_STATUS_TYPE_TRANSLATOR = {
     2: _("unexpected result"),
 }
 
+HEALTH_CHECK_JSON_STATUS_TYPE_TRANSLATOR = {
+    0: "DOWN",
+    1: "OK",
+    2: "WARNING",
+}
+
 
 class HealthCheckException(Exception):
     pass
@@ -44,6 +50,9 @@ class BaseHealthCheckBackend(object):
                 setattr(self, "_status", e.code)
 
         return self._status
+
+    def json_status(self):
+        return HEALTH_CHECK_JSON_STATUS_TYPE_TRANSLATOR[self.status]
 
     def pretty_status(self):
         return "%s" % (HEALTH_CHECK_STATUS_TYPE_TRANSLATOR[self.status])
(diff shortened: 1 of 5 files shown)


health_check/urls.py
```diff
@@ -2,10 +2,11 @@
...
 
 import health_check
-from health_check.views import home
+from health_check.views import home, json_view
 
 health_check.autodiscover()
 
 urlpatterns = [
     url(r'^$', home, name='health_check_home'),
+    url(r'^json/?$', json_view, name='health_check_json_view')
```
codingjoe: What do you think of not add a separate path but making the view header sensitive? You could request the response to be a json.
```
Accept: application/json
```

health_check/backends/base.py
```diff
@@ -15,6 +15,12 @@ class HealthCheckStatusType(object):
     2: _("unexpected result"),
 }
 
+HEALTH_CHECK_JSON_STATUS_TYPE_TRANSLATOR = {
```
codingjoe, line 18: This could be an enum.

tests/test_views.py
```diff
@@ -0,0 +1,44 @@
+from django.test import TestCase
+
+from health_check.plugins import plugin_dir
+from health_check_cache.plugin_health_check import CacheBackend
+
+try:
+    # Django >= 1.4, < 2.0 import
+    from django.core.urlresolvers import reverse
```
codingjoe: No need to add support for any version prior to 1.8

codingjoe (changes requested):
In general I like the idea.
I don't know if it's not better to use literals instead of integers in the JSON.

Besides that, you really should add some documentation here.

health_check/urls.py
aleksihakli: That's a good idea. I'll just add a different view per Accept header and add some documentation :)

health_check/backends/base.py
aleksihakli: Should we refactor the original `HEALTH_CHECK_STATUS_TYPE_TRANSLATOR` as well, then? The current implementation uses the same logic for both; look up pretty response status based on the status code returned by the backend.

aleksihakli:
@codingjoe I refactored the view to return the status with the `Accept: application/json` header and added tests for the JSON output. I'll rebase the commit chain on top of the 2.0 rewrite later to keep the old attributions here.

aleksihakli:
Up-to-date PR in #80.

Issue #70, opened 2016-12-19
Supports for Celery4


Beomi:
Celery4 is relased now.
Does django-health-check supports celery 4?

aleksihakli:
Hey @Beomi!

The current [Celery package we use for testing is actually 4.0.1](https://github.com/KristianOellegaard/django-health-check/blob/master/requirements-dev.txt), so the current version already supports Celery 4.

aleksihakli:
*UPDATE*: On a closer inspection it seems like `@shared_task` in Celery 3 and 4 specific and the Celery support can be checked with the same code base for them:

http://docs.celeryproject.org/en/3.1/django/first-steps-with-django.html?highlight=shared_task
http://docs.celeryproject.org/en/v4.0.2/django/first-steps-with-django.html?highlight=shared_task

But Celery 2 actually uses the `@task` decorator that is available from `celery.task`:

http://docs.celeryproject.org/en/2.5-archived/reference/celery.app.task.html?highlight=app.task#module-celery.app.task

Since Celery 2 is actually the archived Celery version that does not receive any updates, perhaps we should do the following renames for the next release:

- Rename `health_check_celery` to `health_check_celery2` OR deprecate the old health checker AND
- Rename `health_check_celery3` to `health_check_celery` as it supports both Celery 3 and Celery 4. This way it is easier to understand that Celery 3 and 4 are actively supported and Celery 2 is the version that is specially treated.

codingjoe:
As I said in the ticket. Just drop the celery 2 backend.

aleksihakli:
@codingjoe This could be closed.

PR #73, opened 2016-12-23
Rename health_check_celery plugins

aleksihakli:
Update names for packages based on current stable releases, address #70.

- Rename health_check_celery3 to health_check_celery.
- Rename health_check_celery to health_check_celery2.
- Add change notices to README.

Upgrading should be very easy for existing users, it will simply require renaming the imported celery  package in existing applications' INSTALLED_APPS.


codingjoe (changes requested):
This is is a breaking change and requires a new major release. In that case, just drop the celery 2 backend.
Actually we don't even need to support celery 3.

aleksihakli:
Already implemented in #75.

PR #75, opened 2016-12-29
Version 2

codingjoe:
@aleksihakli since we want to do a couple breaking changes, I figured to to them all at once, to easy migration for users.

Let me know what you think of the changes, I'd very much appreciate your feed.


.travis.yml
```diff
@@ -1,42 +1,32 @@
...
 sudo: false
-cache:
-- pip
+dist: trusty
+cache: pip
 python:
-- '2.7'
-- '3.4'
-- '3.5'
-- pypy
```
aleksihakli, line 9: Do we want to drop PyPy support? It is a very popular distribution and just received a lot more funding, so I think keeping it around wouldn't really hurt anyone.

tox.ini
```diff
@@ -1,12 +1,13 @@
 [tox]
-envlist = py{27,34,35}-django{18,19,master},qa
+envlist = py{27,35,36}-dj{18,19,master},qa
 
 [testenv]
 deps=
     -rrequirements-dev.txt
-    django18: Django>=1.8,<1.9
```
aleksihakli: This might make testing slower as PyPI packages are cached and installed very fast in pip caches whereas tarballs and zip installs from GitHub repositories might be very slow to download and won't be cached in any manner.

health_check_celery/tasks.py
```diff
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
-from celery.task import task
+from __future__ import absolute_import
```
aleksihakli: Is `absolute_import` necessary when importing `shared_task` in here?

aleksihakli (approved):

aleksihakli (commented):
Seems mostly good. Some slight details might be looked into, like the PyPy support removal and `__future__` import clauses. Are those necessary with Python 2.7 in here?

Documentation about the new structure should probably be added to README.rst before releasing 2.0. We should also add the `AsyncResult.get()` Celery changes and JSON endpoints so that there is no need for immediate 2.1 and 2.2 releases from those API alterations after releasing 2.0.

I can rebase the JSON endpoint and add the `Accept` header modifications after this is merged.

aleksihakli:
As a final note we should probably consider adding #64 to 2.0 to simplify the redundant logger logic and add better quality tracebacks on errors, if that patchset seems good to you. It's been up for review for quite a bit :)

.travis.yml
codingjoe: That's good for PyPy, but Django officially doesn't support PyPy. Furthermore PyPy currently only implements CPython 2.7 syntax. Django will drop Python 2 support in Django 2.

aleksihakli: Django will actually support Python 2.7 throughout Django 1.11 LTS which should be supported until (at least) April 2020. So all nice Django package authors are stuck with supporting that, too, if we want to be human to people who haven't migrated to Python 3 yet. So, PyPy doesn't really cost anything to support, and the PyPy project might well get PyPy working with Python >= 3.5 well before 2020. They have also talked a lot about supporting Python 2 and Python 3 with the same code base (and the same JIT).

I, too, use PyPy with some projects and it just always sucks when package support is dropped when it doesn't cost anything to keep - if you support Python 2.7, supporting PyPy is always no-effort with existing test infrastructure.

codingjoe:
@aleksihakli I got a little bit more creative. I added a different way of logging that the one you suggested in #64. I made it very similar to the design of form errors in Django. It should now give you the original stack in the logs, but should not expose it to the status page.

Besides that I removed the auto detection. The less magic the better. I went with explicit registration during app.ready.

That allowed me to change the structure a lot and drop tons of code. Lets see what we end up with.

aleksihakli:
@codingjoe The changes seem great, I like the new AppConfig approach. Only note about it is that the [`AppConfig.ready()` method needs to be re-entrant in correct implementations (see the note)](https://docs.djangoproject.com/en/dev/ref/applications/#django.apps.AppConfig.ready) and currently it is not because `plugin_dir.register()` raises on repeated calls.

Otherwise everything seemed good :)

health_check/contrib/celery/plugin_health_check.py
```diff
@@ -21,22 +19,13 @@ def check_status(self):
...
-                sleep(0.5)
-        except IOError:
-            raise ServiceUnavailable("IOError")
-        except:
-            raise ServiceUnavailable("Unknown error")
-
-        raise ServiceUnavailable("Unknown error")
+            result.get(timeout=timeout)
+            if result.result != 8:
+                self.add_error(ServiceReturnedUnexpectedResult("Celery return wrong result"))
```
aleksihakli: Should be `returned`.

aleksihakli (changes requested):
Few small typos in commit name and messages, could maybe be fixed before merging into upstream.

health_check/views.py
```diff
@@ -7,15 +7,15 @@
 
 def home(request):
```
aleksihakli: How about we name this view to `index` and give name it `healthcheck:index` so we won't have multiple different terms about it floating around and have correct URL namespacing for it as well. This way we won't clash with any view names in other applications.

health_check/views.py
```diff
@@ -7,15 +7,15 @@
...
-        if not plugin.status:  # Will return True or None
-            working = False
+        plugin.run_check()
         plugins.append(plugin)
+        errors += plugin.errors
     plugins.sort(key=lambda x: x.identifier())
 
-    if working:
+    if errors:
         return HttpResponse(loader.render_to_string("health_check/dashboard.html", {'plugins': plugins}))
```
aleksihakli: Could rename this to `health_check/index.html` as well, a bit more idiomatic.

codingjoe:
@aleksihakli I updated the documentation and fixed some minor things.

PR #80, opened 2017-01-09
Added support for a JSON health endpoint (rebased)

aleksihakli:
Added support for JSON based health checks with tests on top of the current master branch.

### DIFF
--- a/health_check/views.py
+++ b/health_check/views.py
@@ -22,9 +22,12 @@ class MainView(TemplateView):
         status_code = 500 if errors else 200
 
         if 'application/json' in request.META.get('HTTP_ACCEPT', ''):
-            return JsonResponse(
-                {str(p.identifier()): str(p.pretty_status()) for p in plugins},
-                status=status_code
-            )
+            return self.render_to_response_json(plugins, status_code)
 
         return self.render_to_response({'plugins': plugins}, status=status_code)
+
+    def render_to_response_json(self, plugins, status):
+        return JsonResponse(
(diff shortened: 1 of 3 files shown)


aleksihakli:
@codingjoe I added support for the JSON responses based on the `Accept` header.

I don't know if a CBV is the best option for status rendering as we might want to customize the endpoint and add support for multiple different types of medias such as YAML etc. in the long run. Should we just use FBVs in the endpoint?

aleksihakli:
@codingjoe I think this is ready work-wise and good for merging from my part if it looks good to you.

health_check/views.py
```diff
@@ -19,4 +20,11 @@ def get(self, request, *args, **kwargs):
             errors += plugin.errors
         plugins.sort(key=lambda x: x.identifier())
         status_code = 500 if errors else 200
+
+        if 'application/json' in request.META.get('HTTP_ACCEPT', ''):
+            return JsonResponse(
+                {str(p.identifier()): str(p.pretty_status()) for p in plugins},
```
codingjoe: I would put this part into a different method. That way users can overwrite the structure of the dict that gets serialised into a JSON

codingjoe (changes requested):
@aleksihakli I like it, just one tiny comment.
Oh and please add documentation 😉

aleksihakli:
@codingjoe I refactored the JSON endpoint so that it can be customized easily and added docs about the JSON output and customization to README.

I guess #31 can be closed after this as well, finally :)

codingjoe:
Released in 2.2.0

PR #102, opened 2017-03-15
Fixes expiration issue with timezone

courcelm:
Replace expiration expressed in datetime object with integer seconds to avoid expiration issue (reported as Discarding revoked task).


codingjoe:
Thanks!

Released in 2.2.3

PR #163, opened 2017-12-19
Multiple celery queues

mmagician:
This PR is an enhancement which allows for health check of any number of celery queues.
The requirement is that a user specifies a CELERY_QUEUES setting in their settings file, e.g.:

`CELERY_QUEUES = {
    'default': {
        'exchange': 'default',
        'binding_key': 'default',
        'display_name': 'Celery queue 1'
    },
    'queue2': {
        'exchange': 'queue2',
        'routing_key': 'queue2',
        'display_name': 'Celery queue 2'
    }
}`

If display_name is not specified, the displayed name will default to CeleryHealthCheckDefault or CeleryHealthCheckQueue2 (i.e. using the key of the dictionary)

### DIFF
--- a/health_check/contrib/s3boto3_storage/__init__.py
+++ b/health_check/contrib/s3boto3_storage/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = 'health_check.contrib.s3boto3_storage.apps.HealthCheckConfig'
--- a/health_check/contrib/s3boto3_storage/apps.py
+++ b/health_check/contrib/s3boto3_storage/apps.py
@@ -1,0 +1,11 @@
+from django.apps import AppConfig
+
+from health_check.plugins import plugin_dir
+
+
+class HealthCheckConfig(AppConfig):
+    name = 'health_check.contrib.s3boto3_storage'
+
+    def ready(self):
+        from .backends import S3Boto3StorageHealthCheck
+        plugin_dir.register(S3Boto3StorageHealthCheck)
--- a/health_check/contrib/s3boto3_storage/backends.py
+++ b/health_check/contrib/s3boto3_storage/backends.py
@@ -1,0 +1,24 @@
+import logging
+
+from health_check.storage.backends import StorageHealthCheck
+
+
+class S3Boto3StorageHealthCheck(StorageHealthCheck):
+    """
+    Tests the status of a `S3BotoStorage` file storage backend.
+
+    S3BotoStorage is included in the `django-storages` package
+    and recommended by for example Amazon and Heroku for Django
+    static and media file storage on cloud platforms.
+
+    ``django-storages`` can be found at https://git.io/v1lGx
+    ``S3Boto3Storage`` can be found at
+        https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py
+    """
+
+    logger = logging.getLogger(__name__)
+    storage = 'storages.backends.s3boto3.S3Boto3Storage'
+
+    def check_delete(self, file_name):
+        storage = self.get_storage()
+        storage.delete(file_name)


README.rst
```diff
@@ -51,6 +51,20 @@ Add the ``health_check`` applications to your ``INSTALLED_APPS``:
         'health_check.contrib.s3boto_storage',      # requires boto and S3BotoStorage backend
     ]
 
+If you're using the celery health check, make sure your CELERY_QUEUES are updated (in settings.py).
```
codingjoe: I think you mean `CELERY_TASK_ROUTES`, if in fact your settings prefix is `CELERY` which it doesn't have to be.

I would also recommend to add a link to the celery documentation here.
Also I believe it's `routing_key` not binding key, see http://docs.celeryproject.org/en/latest/userguide/routing.html#how-the-queues-are-defined

health_check/contrib/celery/apps.py
```diff
@@ -8,4 +9,20 @@ class HealthCheckConfig(AppConfig):
 
     def ready(self):
         from .backends import CeleryHealthCheck
-        plugin_dir.register(CeleryHealthCheck)
+
+        # Celery queues defined in SETTINGS as CELERY_QUEUES:
+        for queue, queue_dict in settings.CELERY_QUEUES.items():
```
codingjoe: I would recommend to use `current_app.amqp.queues` as this will return all queues even if you only have task routing defined and implicit queues.

AUTHORS
```diff
@@ -8,3 +8,4 @@ The following people have contributed to django-health-check:
 - Johannes Hoppe
 - Aleksi Häkli
 - Mark Speedy
+- Marcin Górny
```
codingjoe: please add new line to EOF

codingjoe (changes requested):
Hi @mmagician thanks! This looks really promising, very good idea!
My recommendation would be to not base it on a setting but on the actual configured queues.
This can be done using `current_app`.

This way there isn't any manual setup required and it will work for more scenarios.

I looking forward to use this! Again thanks!

README.rst
mmagician: If using CELERY_TASK_ROUTES, the current_app.amqp.queues will not see any queues. And setting.task_queues will not work in Django either (see https://github.com/celery/celery/pull/4467), so in the end I think this should stay as CELERY_QUEUES.
 Regarding the routing_key, yes here I think you're right - I've removed the dictionary settings and just referenced a link.

health_check/contrib/celery/apps.py
mmagician: The implementation of amqp.queues seems to initialise based on CELERY_QUEUES anyway, see (in celery.app.amqp):

```
def queues(self):
        """Queue name⇒ declaration mapping."""
        return self.Queues(self.app.conf.CELERY_QUEUES)
```

I tried it with `current_app.amqp.queues` and while seems to work with my Django app running, it fails on the tests, I guess because these are just python tests and no Django app was launched to initialise it.
Any other suggestions appreciated!

AUTHORS
mmagician: Done.

health_check/contrib/celery/apps.py
codingjoe: Hi @mmagician yes, it works in your tests because you set the setting earlier. For it to actually work in your tests you will need to initiate Celery in the test app.

```python
app = Celery('proj')
app.config_from_object('django.conf:settings', namespace='CELERY')
```
See, http://docs.celeryproject.org/en/latest/django/first-steps-with-django.html

README.rst
codingjoe: If you resolve my [other comment](https://github.com/KristianOellegaard/django-health-check/pull/163#discussion_r160352231) it doesn't matter how you name your prefix. It will handle all. In this case I would just drop this line. Since there are many ways to configure multiple queues.

codingjoe:
@mmagician good work! Don't give up, I am really looking forward to merge this in 👍

mmagician:
Thanks a lot for all the help @codingjoe , I've applied the changes and tested. Should be ready to go!

tests/testapp/__init__.py
```diff
@@ -0,0 +1,5 @@
+from __future__ import absolute_import, unicode_literals
```
codingjoe: rm

tests/testapp/celery.py
```diff
@@ -0,0 +1,6 @@
+from __future__ import absolute_import, unicode_literals
```
codingjoe: rm

health_check/backends.py
```diff
@@ -8,7 +8,7 @@
 logger = logging.getLogger('health-check')
 
 
-class BaseHealthCheckBackend:
+class BaseHealthCheckBackend(object):
```
codingjoe: rm, we don't support Python 2

codingjoe (changes requested):
Great! Almost there! We don't Python 2 tho, please remove all Python 2 related changes. (I added comments to highlight then.)

codingjoe:
Never mind. Did it myself.

Released in 3.2.0

mmagician:
Great, thanks @codingjoe , I was unavailable for the last 2 weeks!
I'm happy to see it released!

PR #167, opened 2018-01-14
Correct URL route entry in README. Fixes #166

### LINKED ISSUE #166: Django system check warning on URL route entry
The current README suggests the following urls.py entry:
```
url(r'^ht/$', include('health_check.urls')),
```
However, running `django-admin check` then yields:
```
System check identified some issues:

WARNINGS:
?: (urls.W001) Your URL pattern '^ht/$' uses include with a regex ending with a '$'. Remove the dollar from the regex to avoid problems including URLs.
```
Removing the dollar sign caused the warning to go away, with no apparent ill effects regarding django-health-check app functionality.

justinmayer:


codingjoe (approved):
Again good catch!

PR #168, opened 2018-01-14
README: Migrations must be run if using DB check

justinmayer:
I was confused when the DB check reported a "Database Error" even though everything seemed to be functioning normally. Upon further investigation, I discovered that there is a migration that must be run for the DB check to function, so I took the liberty of documenting that in the README.


codingjoe (approved):

codingjoe:
Thanks @justinmayer, good catch!

PR #169, opened 2018-01-15
Add checks for disk and memory utilization

justinmayer:
Full disks and out-of-memory conditions are among the primary causes of
service outages, and they are easy to check for.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1,0 +1,32 @@
+Settings
+========
+
+Settings can be configured via the ``HEALTH_CHECK`` dictionary.
+
+``psutil``
+----------
+
+The following default settings will be used to check for disk and memory
+utilization. If you would prefer different thresholds, you can add the dictionary
+below to your Django settings file and adjust the values accordingly. If you want
+to disable any of these checks, set its value to ``None``.
+
+.. code:: python
+
+    HEALTH_CHECK = {
+        'DISK_USAGE_MAX': 90,  # percent
+        'MEMORY_MIN' = 100,    # in MB
+    }
+
+With the above default settings, warnings will be reported when disk utilization
+exceeds 90% or available memory drops below 100 MB.
+
+.. data:: DISK_USAGE_MAX
+
+   Specify the desired disk utilization threshold, in percent. When disk usage
+   exceeds the specified value, a warning will be reported.
+
+.. data:: MEMORY_MIN
+
+   Specify the desired memory utilization threshold, in megabytes. When available
+   memory falls below the specified value, a warning will be reported.
--- a/README.rst
+++ b/README.rst
@@ -4,14 +4,15 @@ django-health-check
 
 |version| |ci| |coverage| |health| |license|
 
-This project checks the health for a number of backends and sees if they are able
-to connect and do a simple action.
+This project checks for various conditions and provides reports when anomalous
+behavior is detected.
 
-The following health check backends are bundled into this project:
+The following health checks are bundled with this project:
 
-- Cache
-- Database
-- Storage
+- cache
+- database
+- storage
+- disk and memory utilization (via ``psutil``)
 - AWS S3 storage
 - Celery task queue
 
@@ -19,10 +20,22 @@ Writing your own custom health checks is also very quick and easy.
 
 We also like contributions, so don't be afraid to make a pull request.
 
+Use Cases
+---------
+
+The primary intended use case is to monitor conditions via HTTP(S), with
+responses available in HTML and JSON formats. When you get back a response that
+includes one or more problems, you can then decide the appropriate course of
+action, which could include generating notifications and/or automating the
+replacement of a failing node with a new one. If you are monitoring health in a
+high-availability environment with a load balancer that returns responses from
+multiple nodes, please note that certain checks (e.g., disk and memory usage)
+will return responses specific to the node selected by the load balancer.
+
 Supported Versions
 ------------------
 
-We officially only support the latest Version of Python as well as the
+We officially only support the latest version of Python as well as the
 latest version of Django and the latest Django LTS version.
 
 .. note:: The latest version to support Python 2 is 2.4.0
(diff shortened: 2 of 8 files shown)


codingjoe:
Hi @justinmayer thanks for the contribution. Thanks for your PRs yesterday too!

I got to be honest, I am not so sure about this particular pull-requests and the two new checks. I would have encouraged you to open a ticket first, to discuss them before you spend all that effort.
They are well written, don't get me wrong. I believe they are conceptually misleading.

See, in a production setup, you would have at least 2+ nodes behind a load balancer. In a multi node environment you can not use local discs, but need to use a storage solution. When it comes to memory, you get a different node, each time you GET the health check endpoint. Since the load balancer would randomly route the request to a different node.

Besides that, the cluster should be able to detect unhealthy nodes and replace them automatically.

I am not sure about adding a feature, that does not support multiple nodes and by definition isn't suited for a high availability setup.

Again, that's just my humble opinion. Let me know if I am missing a point.
Best
-Joe

justinmayer:
Thank you for your comments, Joe. Please allow me to briefly explain in-line.

> See, in a production setup, you would have at least 2+ nodes behind a load balancer.

There are many, **many** production environments that utilize a single node without a load balancer.

> In a multi node environment you can not use local discs, but need to use a storage solution.

Each node presumably runs on a server that has a disk. A disk that can, and will, fill up if not monitored. Happens all the time.

> When it comes to memory, you get a different node, each time you GET the health check endpoint. Since the load balancer would randomly route the request to a different node.

Sure, but wouldn't the app developer want to know if a particular node is running out of memory?

> Besides that, the cluster should be able to detect unhealthy nodes and replace them automatically.

I think that makes a lot of assumptions about the average, everyday production environment that may not hold up to close scrutiny. Yes, there are some high-availability configurations that behave the way you describe. There are also many thousands of environments that do not behave that way. 

> I am not sure about adding a feature, that does not support multiple nodes and by definition isn't suited for a high availability setup.

I'm certainly open to improving these checks so that, if possible, they can be useful in a wide range of environments. But either way, does it not seem overly limiting to say that these features have no merit simply because some environments may not need them?

codingjoe:
@justinmayer I see your point. I work mostly on large scale projects. If you work for an agency and do small sites for clients, I see how you end up on a single machine.
Still none both of the checks support ONLY a single node setup, thus could be misleading. We could merge them still with a proper warning. I personally don't feel good about it tho, since I don't have any commercial interest in maintaining those features.

I let get a third opinion:
@KristianOellegaard @stefanfoulis what do you guys think? Core or 3rd party library?

KristianOellegaard:
I can only speak for our setup - we certainly also use load balancers with a cluster behind it - however, we run the health check not only for the loadbalancer itself but also check each node's healtcheck. In that case it would make sense to look for disk space for us - so I wouldn't mind that particular check. If django-health-check fails on any of our machines 10 times, we replace it with a new one.

However, I would probably not use the RAM check since that could happen under specific heavy loads where you wouldn't want to have the machine killed and replaced automatically.

I noticed it requires psutils, in which case I think it would make sense to put it in the contrib module perhaps?

codingjoe:
Ok, lets do it. @justinmayer please move it to `contrib.psutils` and add a note to the documentation about the possible side effects.

justinmayer:
Excellent. Many thanks, Johannes and Kristian. I will move it to `contrib.psutils` and add some corresponding documentation. Might be a couple of weeks until I can get to it.

justinmayer:
Hi folks. Thanks for your patience and guidance. I updated this pull request with the following improvements:

* moved disk/memory checks into `contrib`
* moved settings into `DJANGO_HEALTH_CHECK` namespaced dictionary
* checks can be disabled by setting to `None`
* warnings include hostname to identify nodes in HA environments
* added use case documentation, including usage in HA environments

I believe this should address the topics raised in our previous discussion. What do you think?

codingjoe (commented):
@justinmayer thats really good work. Before I get into the details, we just added a sphinx docuentation (`python setup.py build_sphinx`).
Would you be so kind to create a new `docs/contrib.rst` file an document the two health checks there including the settings. Please add a second filen `docs/settings`, that docuements the settings in more detail.

justinmayer:
I added the two requested reST files. Running `python setup.py build_sphinx` yields:
```
error: invalid command 'build_sphinx'
```

docs/settings.rst
```diff
@@ -0,0 +1,22 @@
...
+
+The following default settings will be used to check for disk and memory
+utilization. If you would prefer different thresholds, you can add the dictionary
+below to your Django settings file and adjust the values accordingly. If you want
+to disable any of these checks, set its value to ``None``.
+
+.. code:: python
+
+    HEALTH_CHECK = {
+        'DISK_USAGE_MAX': 90,  # percent
```
codingjoe, line 17: In this section it would be good to actually have a headline for each of those keys in the dictionary and briefly explain what they are for.

codingjoe (commented):
Very good work. if you want to test the documentation you will need to install sphinx. I forgot to mention that.

Anyways, I had one last commend. Almost there!

justinmayer:
No worries. Perhaps you might want to add `sphinx` to requirements-dev.txt?

I added a header for each setting along with a brief description, as requested. Looking forward to seeing this merged and included in a new release. 😄

justinmayer:
Hi @codingjoe. Have you had a chance to review the latest changes? Any thoughts?

codingjoe (approved):
Dude, this is pure magic, thanks! <3

codingjoe:
Released in [3.5.0](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.5.0).

Thanks again @justinmayer for your patience, I am really happy with the final result.

justinmayer:
@codingjoe: Very pleased to see this merged and glad that you're happy with it. It was nice collaborating with you. 😁

PR #182, opened 2018-01-28
Fix #178 -- Close connections when exeting threads

### LINKED ISSUE #178: health_check.db in 3.3.0: FATAL:  connection limit exceeded for non-superusers
The latest release (3.3.0) is having an issue for the health_check.db in 3.3.0. The database connections (Postgres in my case) are piling up every time the check is run. The issue is not found in 2.2.1 to 3.2.0.

I guess it is related to this new feature: "Use futures and parallel execution to improve performance"

codingjoe:
Django uses separate connections for each thread be default.
Open connections are only closed by default when the
`request_finished` signal is called, which does not happen in our
case.

### DIFF
--- a/health_check/views.py
+++ b/health_check/views.py
@@ -22,7 +22,11 @@ class MainView(TemplateView):
 
         def _run(plugin):
             plugin.run_check()
-            return plugin.errors
+            try:
+                return plugin.errors
+            finally:
+                from django.db import connection
+                connection.close_if_unusable_or_obsolete()
 
         with ThreadPoolExecutor(max_workers=len(plugins) or 1) as executor:
             for ers in executor.map(_run, plugins):


courcelm:
This fix works if CONN_MAX_AGE = 0 in Django database settings. Otherwise, the connections are still pilling up.

codingjoe:
@courcelm that's a good start I guess. I really don't have any idea how to close a connection forcefully before the connection timeout is reached.

codingjoe:
Released in [3.4.2](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.4.2)

courcelm:
@codingjoe  I would suggest to use close() instead of close_if_unusable_or_obsolete(). This way it will not cause any problem to anyone with CONN_MAX_AGE > 0.  

It seems db health check is not re-using existing connection and makes a new one every time from what I can see. If we want to re-use connection, maybe you should exclude this db check from the parralel execution.

codingjoe:
Ok @courcelm would you mind creating a Pull-Request for that?

PR #188, opened 2018-04-25
`critical` flag, better JSON response

alexdlaird:
- Adding a `critical` flag to identify services that should not cause a health check failure
- Adding better JSON response (to include response time)


alexdlaird:
Closing, as I accidently made this off of `master`. Reopnening as https://github.com/KristianOellegaard/django-health-check/pull/189

PR #189, opened 2018-04-25
`critical` flag, better JSON response

alexdlaird:
- Adding a `critical` flag to identify services that should not cause a health check failure
- Adding better JSON response (to include response time)
- Adding to README code example to reinit a backend so it can be renamed and have its default flags changed (for instance, the new `critical` flag)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 .coverage
 .eggs
 /docs/_build
+.pytest_cache
--- a/README.rst
+++ b/README.rst
@@ -158,6 +158,27 @@ The backend will return a JSON response:
 Optionally, the ``JSON_VERBOSE`` can be set to ``True`` in ``HEALTH_CHECK`` settings to output additional details about
 each plugin (for instance, response time).
 
+Overriding a health check
+-------------------------
+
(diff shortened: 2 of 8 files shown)


codingjoe:
Hi @alexdlaird thanks for the contribution. Seems like you have been quite busy here. That's great! Only problem I have to review it all... could you do me a favor a split things up a little.
I see that you did changes to the usage of concurrent futures as well as to the json view a .gitignore file.
I would suggest to put all of those changes into separate PRs. That way I can merge and release them as quick as possible without being held back by another part of the patch.
Let me know if you need any help.

alexdlaird:
@codingjoe Sorry about that :). Was pushing a bit from my own fork to get it to where I wanted it for my project—I've since tagged that and am using the changes in production, so I'll go back and break this into two (maybe three) different PRs for you to evaluate.

alexdlaird:
Alright @codingjoe, I've created three other PRs for you (added a Twilio health check as well)—I offer them to you as-is. They're a bit cobbled together from my own fork, as I added additional changes to my own fork that are more specific to my use case and I don't think you'd want to merge into the mainline branch, so if they work for you, great! Otherwise feel free to close them out.

PR #190, opened 2018-05-07
Adding a Twilio health check.

alexdlaird:


health_check/contrib/twilio/backends.py
```diff
@@ -0,0 +1,34 @@
...
+                    if component['name'] in self.services and component['status'] != 'operational':
+                        self.add_error(ServiceWarning(
+                            '{} - {}'.format(component['name'], component['status'])))
+            else:
+                response = urllib.request.urlopen('https://gpkpyklzq55q.statuspage.io/api/v2/status.json',
+                                                  timeout=timeout)
+                data = json.loads(response.read().decode('utf-8'))
+                if data['status']['indicator'] != 'none':
+                    self.add_error(ServiceWarning(data['status']['indicator']))
+        except BaseException as e:
```
codingjoe, line 33: This is too broad. Please make sure to narrow down your try block and handle all possible exceptions and provide better exception messages. In this case I would also suggest to just use `raise from`

health_check/contrib/twilio/backends.py
```diff
@@ -0,0 +1,34 @@
...
+        self.services = getattr(self, 'services', [])
+
+    def check_status(self):
+        timeout = getattr(settings, 'HEALTHCHECK_TWILIO_TIMEOUT', 5)
+
+        try:
+            if len(self.services) > 0:
+                response = urllib.request.urlopen('https://gpkpyklzq55q.statuspage.io/api/v2/summary.json',
+                                                  timeout=timeout)
+                data = json.loads(response.read().decode('utf-8'))
```
codingjoe, line 22: `json.load` excepts stream like the response object, so:
```python
json.load(response)
```
does the same thing, at least in Python 3.5 and up

health_check/contrib/twilio/backends.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(self):
+        super(TwilioHealthCheck).__init__()
+        self.services = getattr(self, 'services', [])
+
+    def check_status(self):
+        timeout = getattr(settings, 'HEALTHCHECK_TWILIO_TIMEOUT', 5)
+
+        try:
+            if len(self.services) > 0:
+                response = urllib.request.urlopen('https://gpkpyklzq55q.statuspage.io/api/v2/summary.json',
```
codingjoe, line 20: Please put the url into a CONSTANT (all caps class attribute)

health_check/contrib/twilio/apps.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class HealthCheckConfig(AppConfig):
+    name = 'health_check.contrib.twilio'
+
+    def ready(self):
+        from .backends import TwilioHealthCheck
+
+        if getattr(settings, 'HEALTHCHECK_FANOUT_BACKENDS', True):
+            for service in getattr(settings, 'HEALTHCHECK_TWILIO_SERVICES', ['SMS', 'Phone Numbers']):
```
codingjoe, line 14: Why is this the default and where is it documented?

codingjoe (changes requested):
@alexdlaird thanks for the contribution. That's great, we use twillio too! That being said, I'd like a college of mine to take a look here too. @anneFly 
That being said, I am missing tests as well as documentation. The new settings and the feature itself need to be documented.

Let me know if you need any help! Best-joe

codingjoe:
@alexdlaird oh I see you even work at twillio that great. We already have a health check for you guys but haven't published it yet. My college knows more, lets combine efforts here.

health_check/contrib/twilio/apps.py
```diff
@@ -0,0 +1,20 @@
...
+                celery_class_name = '{}:{}'.format(TwilioHealthCheck.__name__, service.replace(' ', ''))
+
+                celery_class = type(celery_class_name, (TwilioHealthCheck,), {'queues': [service]})
+                plugin_dir.register(celery_class)
```
anneFly, line 18: hmm... it might be that just don't understand completely what's going on here but this whole block looks a bit like a bad copy paste to me. If I understand you're intentions correctly your're registering different plugins for different Twilio Services, e.g. one for "SMS", one for "Phone Numbers", correct? You should change the naming then to not use `celery` because this has nothing to do with celery if I get it correctly. Also: shouldn't it be `{'service': [service]})` instead of `{'queues': [service]})`? Because you're using `self.services = getattr(self, 'services', [])` later in the base class.
Personally, I agree with @codingjoe, I would not set a default here for the services you want to check. That's not intuitive. Also it feels like one setting is enough to configure the health check for different services instead of `HEALTHCHECK_FANOUT_BACKENDS` and `HEALTHCHECK_TWILIO_SERVICES`. The latter one should be enough. If the setting is not configured, just register the normal `TwilioHealthCheck`.

anneFly (commented):
First of all I agree with @codingjoe that this would need tests as well as documentation.

Apart from that I'll just add my thoughts here:
It's an interesting approach. You're querying the Twilio status page which contains the status of each individual service like "Programmable Voice", "Programmable Chat", "Speech Recognition" etc. My thoughts here are: how up to date is the status page actually? I cannot say for sure how the Twilio status page behaves but from my experience status pages are usually a bit behind the real time status. The API might be down for 10 minutes until the status page reflects this. 
Another thought: since Twilio is operating world wide, a lot of notifications about degraded quality are often only relevant for a certain location, e.g. the health check might show that "Programmable SMS" is not working correctly because SMS delivery to a certain provider in Brazil doesn't work atm. This would be rather irrelevant for me sitting in Germany. Should this be a reason for the health check to fail? I'm not sure about the intensions of when the health check is supposed to fail. What do you think @codingjoe ?

The way we currently do our Twilio health check is rather simple. We have a ht plugin that uses the Twilio python client to make a request to Twilio's REST API (simple request listing applications). That's more or less just a check if Twilio's API is not down, nothing fancy. Of course this won't tell you the status of all of Twilio's products, e.g. if you want the health check to fail whenever Twilio's docs are down, this approach  won't work.

alexdlaird:
@anneFly That would be another approach—while you'd get a more prompt result, it would be less fine, so that's the tradeoff. Just depends what you're looking for, I suppose, but ultimately a result saying the Twilio API is up is pretty detached from any of the services, unless you also check the individual for a 5xx response and/or complete a loop with a service (i.e. send a text message, verify the text message was sent, similar to how you're doing the S3 check).

Again, this is use case specific, I suppose. In my use case, Twilio is a non-critical service (see my other PR), thus a delayed (only by some seconds/minutes at most, as StatusPage is programmatically updated) but more reliable/verbose response is better.

health_check/contrib/twilio/apps.py
alexdlaird: Yup, copy/paste error. Sorry about that. Like I said, I was trying to cobble this PR together from my own branch, which more uniquely matches my use case. If I get some time, perhaps I will try to broaden out/add more testing for these PRs to make them more universally applicable if no one else gets to them first.

codingjoe:
Thx @anneFly for the prompt response.

@alexdlaird I like the idea of basing it on the status page over a simple API call. If API calls fail, this information would be end up on a service like Sentry anyways. On the other hand if SMS are not delivered or delayed, is an information only available via the API.

Still tests and docs are a must I also left you a couple comments ealier. Once you address those I am happy to move this ahead.

Best -J

alexdlaird:
@codingjoe Going to close this out, as it sounds like my use case for this at least is too specialized for the mainline branch.

PR #191, opened 2018-05-07
Adding critical flag.

alexdlaird:


codingjoe (changes requested):
Hi @alexdlaird wow, you are busy. That's great.
I really like the idea for the feature, but I would suggest some changes in implementation. Mainly I would add the possibility to add warnings, similar to the `add_error` method.

So a checks isn't critical or not, but it's results are. Does that make sense?

alexdlaird:
@codingjoe I would disagree with that approach, at least in my use-case. I think adding the add_warning similar to add_error is a fine idea, but the `critical` flag would still be necessary, largely because whether or not a test is critical is defined by the use case, not the service. MySQL might be a critical service to you but not to me, so letting the service define itself whether it's critical is not flexible enough (in my use case, I wanted this flexibility with the S3 health check).

If you agree with this, I can refactor to utilize add_warning, then whether or not add_error or add_warning is used will be determined by the state of the services `critical` flag rather than in the view as it is now.

codingjoe:
> @codingjoe I would disagree with that approach, at least in my use-case. I think adding the add_warning similar to add_error is a fine idea, but the critical flag would still be necessary, largely because whether or not a test is critical is defined by the use case, not the service. MySQL might be a critical service to you but not to me, so letting the service define itself whether it's critical is not flexible enough (in my use case, I wanted this flexibility with the S3 health check).

I see, good point. I would still want the fronend to show the test as down, but maybe not return a 500.

> If you agree with this, I can refactor to utilize add_warning, then whether or not add_error or add_warning is used will be determined by the state of the services critical flag rather than in the view as it is now.

I believe we can and should have both. I would motivate you to do both if you introduce a new concept of a warning. You think you are up to this?

alexdlaird:
@codingjoe Changes implemented, with tests.

health_check/backends.py
```diff
@@ -45,9 +46,26 @@ def add_error(self, error, cause=None):
...
+        if isinstance(cause, BaseException):
+            logger.exception(str(warning))
+        else:
+            logger.warning(str(warning))
+        self.warnings.append(warning)
+
     def pretty_status(self):
         if self.errors:
             return "\n".join(str(e) for e in self.errors)
+        elif self.warnings:
```
codingjoe: You will omit warnings if there are errors. I would show them all. Just merge two list ;)

health_check/backends.py
```diff
@@ -11,6 +11,9 @@
 class BaseHealthCheckBackend:
     def __init__(self):
         self.errors = []
+        self.warnings = []
+        self.critical = getattr(self, 'critical', True)
```
codingjoe: ? I can't follow you ? Why would you do that?

health_check/backends.py
```diff
@@ -11,6 +11,9 @@
...
+        self.description = getattr(self, 'description', '')
```
codingjoe: Same here....

health_check/backends.py
```diff
@@ -43,9 +46,26 @@ def add_error(self, error, cause=None):
...
+        if isinstance(warning, HealthCheckException):
+            pass
+        elif isinstance(warning, str):
+            msg = warning
+            warning = HealthCheckException(msg)
+        else:
+            msg = _("unknown error")
+            warning = HealthCheckException(msg)
+        if isinstance(cause, BaseException):
+            logger.exception(str(warning))
```
codingjoe: I don't know if you really want to log at exception level here. I would go for warning.

health_check/backends.py
```diff
@@ -43,9 +46,26 @@ def add_error(self, error, cause=None):
             logger.error(str(error))
         self.errors.append(error)
 
-    def pretty_status(self):
+    def add_warning(self, warning, cause=None):
```
codingjoe: I am sure you can reuse some code here. I would advice to change the structure of the `errors` list and have a new Error type that includes a severity level. Similar to logs really.

Those methods would be merely helpers then.

health_check/plugins.py
```diff
@@ -21,5 +21,12 @@ def register(self, plugin, **options):
         # Instantiate the admin class to save in the registry
         self._registry.append((plugin, options))
 
+    def reregister(self, replace_class, new_plugin, **options):
```
codingjoe: How's this related?

health_check/views.py
```diff
@@ -46,3 +41,11 @@ def render_to_response_json(self, plugins, status):
             {str(p.identifier()): str(p.pretty_status()) for p in plugins},
             status=status
         )
+
+    def _run(self, plugin):
```
codingjoe: self is unused and this way you will run into the GIL since you all threads need to access a thread safe object

codingjoe (changes requested):
I would encourage you to not use multiple lists to introduce severity levels. I would go for a single list of objects, that are ware. Simplest would be a list of tuples or you build your own objects or namedtuples. Objects are probably simplest because you can add str and json repr methods.

health_check/backends.py
alexdlaird: https://github.com/KristianOellegaard/django-health-check/pull/191/files#diff-88b99bb28683bd5b7e3a204826ead112R177

Simply, it determines if that attribute has been set on the object, since that value would be preferred if so, otherwise it sets it to the default.

health_check/backends.py
alexdlaird: Same as above. Just ensuring it's not None.

health_check/plugins.py
alexdlaird: https://github.com/KristianOellegaard/django-health-check/pull/191/files#diff-88b99bb28683bd5b7e3a204826ead112R177

health_check/backends.py
alexdlaird: I don't think I would change the structure of the `errors` list given it's public, and people (like me :P) may be relying on it—I wouldn't want to introduce a breaking change. However, I can easily make the two methods share the same code, so I'll update that.

health_check/plugins.py
alexdlaird: Alternatively, we could add a sort of `def update_registry(self, class, option, new_value):`. Either way, we need a way to override the default value's of plugin attributes (critical flag, description, identifier, etc.)

health_check/backends.py
alexdlaird: Oh, I think I misunderstood what you meant. Yes, I'll just a new "Warning" type of exception and wrap these in that.

alexdlaird: Alright, gonna take the kiddo to the museum now, so done for the weekend. But here's my update: given all the changes that have made at this point, I instead just opted to remove the `warnings` list altogether, as it is now functionally identical to the `errors` list and we're instead managing that's behavior and output with flags, which I think is better anyway.

README.rst
```diff
@@ -165,16 +187,29 @@ Writing a health check is quick and easy:
     from health_check.backends import BaseHealthCheckBackend
 
     class MyHealthCheckBackend(BaseHealthCheckBackend):
+        def __init__(self):
+            super().__init__()
+
+            # Display name on the endpoint.
+            self.identifier = 'My HealthCheck'
```
codingjoe: You are overriding a method name here.

README.rst
```diff
@@ -165,16 +187,29 @@ Writing a health check is quick and easy:
...
+
+            # A useful description of the health check's purpose, which
+            # can be when the status is returned.
+            self.description = 'An awesome and useful health check!'
```
codingjoe: I would suggest to make this a class not instance attribute.
https://www.python-course.eu/python3_class_and_instance_attributes.php

README.rst
```diff
@@ -165,16 +187,29 @@ Writing a health check is quick and easy:
...
+            self.critical = False
+
         def check_status(self):
             # The test code goes here.
             # You can use `self.add_error` or
             # raise a `HealthCheckException`,
             # similar to Django's form validation.
             pass
 
-        def identifier(self):
```
codingjoe, line 175: Please leave as is.

README.rst
```diff
@@ -155,6 +157,26 @@ The backend will return a JSON response:
         "S3BotoStorageHealthCheck": "working"
     }
 
+Overriding a health check
```
codingjoe: I don't thing this is the way it should be done. Simply don't register them in the first place. We only register the plugins if the apps are added.

I am happy to clarify this better in the docs but it's out of scope for this PR.

codingjoe: Please drop the entire section.

health_check/backends.py
```diff
@@ -11,6 +11,9 @@
 class BaseHealthCheckBackend:
     def __init__(self):
         self.errors = []
+        self._identifier = getattr(self, '_identifier', self.__class__.__name__)
```
codingjoe: Unrelated, please leave as is.

health_check/backends.py
```diff
@@ -11,6 +11,9 @@
...
+        self.critical = getattr(self, 'critical', True)
```
codingjoe: Please just use a class attribute. There is not need to overwrite if on each instance. On the contrary, this adds unwanted side effects.
You can no longer overwrite the class attribute on an instance because it's shadowed. That that someone should do that, but it changes behavior.

Besides I don't see the relation to the feature you want to add.

health_check/plugins.py
codingjoe: I wrote a comment further up how to not register plugins to begin with. There is no need for re `update` function.

health_check/views.py
```diff
@@ -29,8 +29,9 @@ def _run(plugin):
                 connection.close()
 
         with ThreadPoolExecutor(max_workers=len(plugins) or 1) as executor:
-            for ers in executor.map(_run, plugins):
-                errors.extend(ers)
+            for plugin, ers in zip(plugins, executor.map(_run, plugins)):
+                if plugin.critical:
```
codingjoe, line 33: This is not what I was going for. Now you just ignore errors if they are not critical. So they will be not shown.

codingjoe (commented):
Hi @alexdlaird please understand that reviewing pull-requests eats up a lot of my time. I want to be as helpful as I can, but it is tricky if you add a lot of changes that are unrelated to your feature request.

Please read my notes and drop everything that is not exclusively related to your feature request. There is another time and place (other PR) for those things.

If you have further questions or are stuck let me know. I am happy to help out if I can.

health_check/views.py
alexdlaird: Correct, as the `errors` list here is only used to determine if we're going to return 500 or 200, and we discussed above that non-critical plugins would would output there errors (see pretty_status method) but not cause the entire endpoint to 500.

codingjoe:
Hi @alexdlaird I took the liberty to propose my own take on the whole issue. Feel free to review #193. If you are happy with the changes will release it asap. Best -Joe

alexdlaird:
Works for me. I already have the necessary changes in my own fork that I'm using, so what you want to put in the mainline is up to you.

Closing PR in favor or #193.

PR #192, opened 2018-05-07
Better json

alexdlaird:
This PR depends on the `critical` flag. Merge that PR first (if you approve).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -155,6 +155,26 @@ The backend will return a JSON response:
         "S3BotoStorageHealthCheck": "working"
     }
 
+Overriding a health check
+-------------------------
+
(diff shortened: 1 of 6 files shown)


codingjoe (changes requested):
There is still loads of stuff in here – like critical or not – that doesn't belong here. I am also not so sure about the setting.
I would rather suggest to add a separate URL or GET parameter in addition to the header.

alexdlaird:
@codingjoe FYI, probably going to revert most of this PR and just add a simple example (illustrating overriding the JSON method in the view, since you already provide that leverage) to the README instead. Sound good?

codingjoe:
sounds good, maybe you want to wait until your other PR is merged?

PR #193, opened 2018-05-24
Add support for warnings that do not error endpoints

codingjoe:
Add a new concept of warnings and non-critical services. Both
will not cause the template and JSON endpoint to response with a
500 status code but with a regular 200. This change is backwards
compatible because the existing warnings are treated as errors by
default.

Close #191

### DIFF
--- a/health_check/contrib/psutil/backends.py
+++ b/health_check/contrib/psutil/backends.py
@@ -2,21 +2,17 @@ import locale
 import socket
 
 import psutil
-from django.conf import settings
 
 from health_check.backends import BaseHealthCheckBackend
+from health_check.conf import HEALTH_CHECK
 from health_check.exceptions import (
     ServiceReturnedUnexpectedResult, ServiceWarning
 )
 
 host = socket.gethostname()
 
-if hasattr(settings, 'HEALTH_CHECK'):
-    DISK_USAGE_MAX = settings.HEALTH_CHECK.get('DISK_USAGE_MAX', 90)  # in %
-    MEMORY_MIN = settings.HEALTH_CHECK.get('MEMORY_MIN', 100)  # in MB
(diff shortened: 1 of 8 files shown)


health_check/views.py
```diff
@@ -23,14 +25,20 @@ def get(self, request, *args, **kwargs):
         def _run(plugin):
             plugin.run_check()
             try:
-                return plugin.errors
+                return plugin.errors, plugin
```
alexdlaird: Is there a reason we don't just return ``plugin`` here? Either way we're changing the method signature, but given it's declared within the scope of the view itself, no one else should be relying on it. Seems cleaner that way.

health_check/contrib/psutil/backends.py
```diff
@@ -2,21 +2,17 @@
...
 
 from health_check.backends import BaseHealthCheckBackend
+from health_check.conf import HEALTH_CHECK
 from health_check.exceptions import (
     ServiceReturnedUnexpectedResult, ServiceWarning
 )
 
 host = socket.gethostname()
 
-if hasattr(settings, 'HEALTH_CHECK'):
```
alexdlaird, line 14: Much better, these were getting ugly.

alexdlaird (approved):

health_check/views.py
codingjoe: You're absolutely right, thanks!

health_check/contrib/psutil/backends.py
codingjoe: Yes, it has outlived it's design.

codingjoe:
Released in [3.6.0](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.6.0)

PR #195, opened 2018-06-08
Add connection error exception for cache backend

Daiern:
Just wondering if it would be a good idea to catch a connection error. I ran into an issue where it would spit back a malformed response because the exception wasn't being caught. Could probably also be added to the other checks if it's accepted here.


codingjoe (approved):

codingjoe:
Released in [3.6.1](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.6.1)

thx @Daiern 👍

PR #198, opened 2018-09-03
Fix typo in an example in the readme

EvgeneOskin:


codingjoe:
Thanks @EvgeneOskin good catch!

PR #201, opened 2018-10-13
Add a NotImplementedError warning to celery health check.

jrpotter:
This can be triggered if the result backend isn't set which unfortunately took a long time for me to sort out because the error message was so ambiguous.

### DIFF
--- a/health_check/contrib/celery/backends.py
+++ b/health_check/contrib/celery/backends.py
@@ -24,6 +24,6 @@ class CeleryHealthCheck(BaseHealthCheckBackend):
         except IOError as e:
             self.add_error(ServiceUnavailable("IOError"), e)
         except NotImplementedError as e:
-            self.add_error(ServiceUnavailable("NotImplementedError: Have you set your CELERY_RESULT_BACKEND?"), e)
+            self.add_error(ServiceUnavailable("NotImplementedError: Make sure CELERY_RESULT_BACKEND is set"), e)
         except BaseException as e:
             self.add_error(ServiceUnavailable("Unknown error"), e)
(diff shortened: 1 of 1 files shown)


health_check/contrib/celery/backends.py
```diff
@@ -23,5 +23,7 @@ def check_status(self):
                 self.add_error(ServiceReturnedUnexpectedResult("Celery returned wrong result"))
         except IOError as e:
             self.add_error(ServiceUnavailable("IOError"), e)
+        except NotImplementedError as e:
+            self.add_error(ServiceUnavailable("NotImplementedError"), e)
```
codingjoe: Could you add a more descriptive message? I know the other exceptions aren't handled much better, but this is a good place to start ;)

codingjoe (commented):
Hi @jrpotter thanks for the contribution. I had one small comment. Let me know when you updated the PR. I am happy to release it right away. Best -joe

jrpotter:
Thanks Johannes; added a bit more of a description. Not sure if that's more along the lines of what you  were looking for though. PTAL!

health_check/contrib/celery/backends.py
```diff
@@ -23,5 +23,7 @@ def check_status(self):
                 self.add_error(ServiceReturnedUnexpectedResult("Celery returned wrong result"))
         except IOError as e:
             self.add_error(ServiceUnavailable("IOError"), e)
+        except NotImplementedError as e:
+            self.add_error(ServiceUnavailable("NotImplementedError: Have you set your CELERY_RESULT_BACKEND?"), e)
```
codingjoe:
```suggestion
            self.add_error(ServiceUnavailable("NotImplementedError: Make sure CELERY_RESULT_BACKEND is set"), e)
```

codingjoe:
Released in [3.8.0](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.8.0)

PR #206, opened 2018-11-25
Allow overriding critical_service in all checks

GabLeRoux:
* Adds settings for all existing checks
* All defaults to critical services
* Display wether the service is critical or not in the html response

Example with following settings:

```python
HEALTH_CHECK = {
    'CRITICAL_STORAGE': False,
    'CRITICAL_DISK_USAGE': False,
    'CRITICAL_MEMORY_USAGE': False,
}

INSTALLED_APPS = [
# ...
    'health_check',
    'health_check.db',
    'health_check.storage',
    'health_check.contrib.psutil',
]
```
image


GabLeRoux:
This PR is currently failing the QA check:
https://travis-ci.org/KristianOellegaard/django-health-check/jobs/459454246

But I'm not quite sure why. Anyway, I wanted some of the plugins, but I wanted a way to mark some of them as non critical.

codingjoe (commented):
I am not so sure about this. I don't know if this should really be configurable or if behavior should be change via inheritance. I would currently opt for the latter and add proper documentation on how to do that. What do you think?

GabLeRoux:
Yes, this is in fact the first thing I wanted to do, but inheritance required me to in create an app for each of them and it was a bit more work to do on my side at first without documentation.

I think that displaying the flag on the page can still be handy.
Do you have ideas on what would be needed for the smallest way to extend and use them?

I was thinking of something like that (draft):

`apps/custom_health_check_db/apps.py`
```python
from django.apps import AppConfig

from health_check.plugins import plugin_dir


class HealthCheckConfig(AppConfig):
    name = 'custom_health_check_db'

    def ready(self):
        from .backends import DatabaseBackend
        plugin_dir.register(DatabaseBackend)
```

`apps/health_check_db/backends.py`
```python
from health_check.db.backends import DatabaseBackend


class DatabaseBackend(DatabaseBackend):
    critical_service = False  # customized value
```

`conf/settings.py`
```python
DJANGO_APPS = [
  # ...
  'health_check',
  'custom_health_check_db',
  # ...
]
```
I did not try it, but the above would make the database non critical service. I think it would still use the same models and db tables, etc.

codingjoe:
I agree, displaying it would be handy.

About the sample:  You don't need to have a separate app. You can just create a new check and register it. The `INSTALLED_APPS` part is really only for convenience.

seanpar203:
I'm closing this as it's very old and doesn't seem specific to problems the package has.

It's nice for configuration but not really a problem.

PR #207, opened 2018-11-25
Fix a code comment that was misleading concenring critical services

GabLeRoux:
I noticed this comment which I think was wrong.

* If the service is critical, it's returning 500 when the check fails.
* If the service is not critical, it's returning 200 even when the check fails

I confirmed this as part of #206 ;)


codingjoe (approved):
Good catch thanks @GabLeRoux

PR #208, opened 2019-01-03
Added a JSON-first view

bantalon:
- The default view is still the same. It is HTML-first, meaning it returns HTML by default and can return JSON based on the Accept header.
- There is a new JSON-first view which returns JSON by default and can return HTML based on the Accept header.

---
This change is [Reviewable](https://reviewable.io/reviews/kristianoellegaard/django-health-check/208)

### DIFF
--- a/health_check/views.py
+++ b/health_check/views.py
@@ -1,6 +1,7 @@
+import copy
+
 from abc import ABC, abstractmethod
 
-import copy
 from concurrent.futures import ThreadPoolExecutor
 
 from django.http import JsonResponse
--- a/health_check/views.py
+++ b/health_check/views.py
@@ -1,3 +1,5 @@
+from abc import ABC, abstractmethod
+
 import copy
 from concurrent.futures import ThreadPoolExecutor
 
@@ -10,11 +12,18 @@ from health_check.exceptions import ServiceWarning
 from health_check.plugins import plugin_dir
 
 
-class MainView(TemplateView):
-    template_name = 'health_check/index.html'
+class Status(object):
+    def __init__(self, plugins, status_code):
+        self.plugins = plugins
+        self.status_code = status_code
 
-    @never_cache
-    def get(self, request, *args, **kwargs):
+    def get_context(self):
+        return {'plugins': self.plugins, 'status_code': self.status_code}
+
+
+class AbstractMainView(ABC, TemplateView):
+    @staticmethod
+    def _get_status():
         errors = []
 
         plugins = sorted((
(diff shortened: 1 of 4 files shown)


codingjoe (commented):
Hi @bantalon, thanks for the contribution. Much appreciated and I like the idea. Might I make a suggestion tho? How about just adding a simple `json=1` GET keyword argument to the URL? That could be implemented, which just a couple lines. Furthermore it is not an uncommon design choice and implemented in DRF too, if I am not mistaken.

Best
-Joe

bantalon:
Hey @codingjoe,

The requirement I am trying to settle: having an health API endpoint which returns the health status in the API default content type - JSON. This must happen even if the client did not specify it needs JSON content in any explicit way. Therefore, I think your suggestion is valid but it orthogonal to my requirement - it allows explicitly "asking" for JSON in another way than using the currently supported `Accept` header.
By the way, can you refer to the relevant DRF documentation?

Best,
Alon

codingjoe:
Hi @bantalon,

> The requirement I am trying to settle: having an health API endpoint which returns the health status in the API default content type - JSON. This must happen even if the client did not specify it needs JSON content in any explicit way. Therefore, I think your suggestion is valid but it orthogonal to my requirement - it allows explicitly "asking" for JSON in another way than using the currently supported Accept header.

So it does solve your problem, that's a good start. Let me try to explain why I made the suggestion.

Technically they are very similar:
```
GET /ht/json/ HTTP/1.1
```
vs
```
GET /ht/?json=1 HTTP/1.1
```
or
```
GET /ht/ HTTP/1.1
{
    "hello": "world"
}
```

Content negotiation is specified in HTTP via request headers, see also [RFC 7231 5.3](https://tools.ietf.org/html/rfc7231#section-5.3).
So it is expected that a request of the same URL can return multiple different content types. This should be the default, as it currently is.

Now, I understand, that you can not always provide the header. In this case, providing a GET parameter as part of the URL seems a valid fallback strategy.

See, I suggested the GET parameter solution, because it does not violate RFC 7231 section 5.3 yet solves your problem. Furthermore it is recognizable for other users from famous packages like DRF.

As a maintainer I need to do the balance act between solving your specific problem and while keeping the bigger picture and health of the package in mind. I hope you understand.

> By the way, can you refer to the relevant DRF documentation?

Not the documentation, but the code, for it, yes:
https://github.com/encode/django-rest-framework/blob/7749e4e3bed56e0f3e7775b0b1158d300964f6c0/rest_framework/negotiation.py#L42-L44

DRF has this concept of content negotiators. An overengineered strategy pattern, if you will so.

bantalon:
The issue is not that my client cannot state the required content type. Rather, my existing server code (not health check related) defaults to json content type when `Accept` header is not specified and I want the new health endpoint to conform to the same default as the rest of my server. Therefore, I am interested in a way to instanciate a view which defaults to json when the client states nothing in the request (neither header nor parameter).
I hope it clarifies the requirement.
If you think it is useful for others as well, I will be happy to get the code in.

codingjoe:
@bantalon I got it. I would suggest adding a method, that handles content negotiation on the view. By default it should service HTML. You can than override this method in your application and customize it the way you want it to behave. Sounds good?

codingjoe:
Since there hasn't been an update for two month I am closing this. Let me know if you want to continue to work on this.

PR #209, opened 2019-01-06
created new health check for celery (celery ping)

witold-gren:
Hi, your package for testing health app is awesome :) 
but some apps have a problem when testing celery module.
When app creates a lot of tasks in queue then testing celery by creating new task always return fault code. This extension give us possibility check that celery app work (i usage celery ping for that).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -70,6 +70,7 @@ Add the ``health_check`` applications to your ``INSTALLED_APPS``:
         'health_check.cache',
         'health_check.storage',
         'health_check.contrib.celery',              # requires celery
+        'health_check.contrib.celery_ping',         # requires celery
         'health_check.contrib.psutil',              # disk and memory utilization; requires psutil
         'health_check.contrib.s3boto_storage',      # requires boto and S3BotoStorage backend
         'health_check.contrib.rabbitmq',            # requires RabbitMQ broker
--- a/health_check/contrib/celery_ping/__init__.py
+++ b/health_check/contrib/celery_ping/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = 'health_check.contrib.celery_ping.apps.HealthCheckConfig'
--- a/health_check/contrib/celery_ping/apps.py
+++ b/health_check/contrib/celery_ping/apps.py
@@ -1,0 +1,12 @@
+from django.apps import AppConfig
+
+from health_check.plugins import plugin_dir
+
+
+class HealthCheckConfig(AppConfig):
+    name = 'health_check.contrib.celery_ping'
+
+    def ready(self):
+        from .backends import CeleryPingHealthCheck
+
+        plugin_dir.register(CeleryPingHealthCheck)
--- a/health_check/contrib/celery_ping/backends.py
+++ b/health_check/contrib/celery_ping/backends.py
@@ -1,12 +1,8 @@
 from django.conf import settings
 
+from celery.app import default_app as app
 from health_check.backends import BaseHealthCheckBackend
 from health_check.exceptions import ServiceUnavailable
-
-try:
-    from celery.app import default_app as app
-except ImportError:
-    from celery import current_app as app
 
 
 class CeleryPingHealthCheck(BaseHealthCheckBackend):
--- a/health_check/contrib/celery_ping/backends.py
+++ b/health_check/contrib/celery_ping/backends.py
@@ -1,6 +1,6 @@
+from celery.app import default_app as app
 from django.conf import settings
 
-from celery.app import default_app as app
 from health_check.backends import BaseHealthCheckBackend
 from health_check.exceptions import ServiceUnavailable
 
--- a/health_check/contrib/celery_ping/backends.py
+++ b/health_check/contrib/celery_ping/backends.py
@@ -1,0 +1,33 @@
+from django.conf import settings
+
+try:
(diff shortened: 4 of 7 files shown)


codingjoe:
@witold-gren thanks for the contribution. This is an interesting approach. I honestly don't want to have/maintain 2 different health checks for celery. Which one do you think is more superior?

peterfarrell:
@codingjoe I think this PR is superior to the current celery check in the project. The current health check for celery only checks to see if a task completes and assumes there is only 1 host with workers running. In many of our my apps, I have workers running on several hosts/vms. So the current check could possibly indicate everything is ok because at least on set of workers is working, but others are down.

peterfarrell:
The ping command from Celery actually checks all registered workers:

```
>>> p = celery_app.control.ping()
>>> p
[{'worker1@server01.example.com': {'ok': 'pong'}}, {'worker1@server02.example.com': {'ok': 'pong'}}]
>>> 
```

codingjoe:
Hi @witold-gren I see. I like the approach, but this would only make sense if you actually check the responses. You would need to check if a worker does not pong back, even though it should.

If you do that, I agree that this approach is superior and should replace the current implementation.

Best
-Joe

peterfarrell:
@witold-gren / @codingjoe,

If no workers are registered, the list returned by `ping`() is empty.

I think you would have to do two checks:

* Check that at least one worker responds
* Check that all workers that do respond have a response of `pong`

 So it makes to me that if you use this particular health-check that at least 1 worker is running.  Since it is possible that the number of workers running will vary if you do scaling, it does not make sense to check that specific number of workers are running -- just that at least one worker is running.

codingjoe:
@peterfarrell why do you you think it is ok, to only have one worker respond? What information would that give you? That at least one worker is actually having no IO issues. Wouldn't you want to know, that all your workers respond? Furthermore, is there a scenario where they don't respond `{'ok': 'pong'}`?

peterfarrell:
@codingjoe In cloud environments or where you horizontally scale workers, you can't make the assumption that there will be a fixed number of workers that respond. Depending on load, I've seen anywhere from 2 - xx workers depending on scaling. Usually the minimum should be at least two if you looking for high availability. In smaller apps, I could see possibly expected a fixed number of workers.

Maybe a configuration settings like `CELERY_MINIMUM_EXPECTED_WORKERS = 2` which would default to 1 if not present. I think this would cover the majority of the use cases. Thoughts?

peterfarrell:
@codingjoe The only thing you can expect is a `pong`.  Celery ping is basically a check to see what workers are alive and responding:

https://github.com/celery/celery/blob/b2668607c909c61becd151905b4525190c19ff4a/docs/userguide/workers.rst#ping

If a worker is not alive or does not respond in the default 1 second timeout, it does not show up in the list of alive workers.

codingjoe:
Hi @peterfarrell, I think I didn't communicate my thoughts well. Of course you don't know how "should" be there, as in load dependent. What you do know tho, is how many workers are in the pool and that all the workers in the pool should respond.

Check out how flower determines the list of workers, currently connected:
https://github.com/mher/flower/blob/c2f9bf6cc6950fc7fc9088675945a4cc0c71d842/flower/views/dashboard.py#L23-L50

Maybe we can combine those too. Or at least make sure that every queue (or a explicit list of queues) have an active consumer.

peterfarrell:
Hi @codingjoe,

Flower doesn't list the worker if it doesn't respond. They are just absent.

It hard codes `worker.alive` without any other logic to set a "failed" status. Flower is functionally doing the same thing as `ping` but also they want the event count return which is why they don't use ping.

https://github.com/mher/flower/blob/c2f9bf6cc6950fc7fc9088675945a4cc0c71d842/flower/views/dashboard.py#L44

I think we at the point of trying to find a silver bullet here.  Agile wise, the PR is good and an improvement to the current approach in master. After the PR is pulled in, incremental improvement can be made from there (like add min / max expected workers).

codingjoe:
> I think we at the point of trying to find a silver bullet here. Agile wise, the PR is good and an improvement to the current approach in master. After the PR is pulled in, incremental improvement can be made from there (like add min / max expected workers).

@peterfarrell you have a point. I also don't want to overcomplicate things. On the other hand, this package has many users. Getting it right, is somewhat important. Anyhow, how about we do at least two things if possible:
1. we test that `ping_result` isn't empty
2. we check that all queues that are configured have at least one listener

maszaa:
When will this check be merged? I'm having problems with the current Celery check as at times all my workers are busy (can't handle anymore tasks) but that doesn't mean they are dead/failed.

maszaa:
I've forked the repo and rebased the branch on top of master and all tests pass: https://github.com/maszaa/django-health-check/tree/celery-ping

codingjoe:
@maszaa if you have a patch that improved the problem I noted earlier, feel free to submit it as a pull-request. I am happy to merge it, once all issues are resolved. Best Joe

maszaa:
So ping response should be checked that at least one worker for each configured queue responds with 'pong'?

codingjoe:
> So ping response should be checked that at least one worker for each configured queue responds with 'pong'?

I guess that's the idea here.

maszaa:
Merged on https://github.com/KristianOellegaard/django-health-check/pull/272

PR #214, opened 2019-04-13
Travis: Fix deploy distribution setting

JayH5:
This setting is being ignored because it is in the wrong place. This means that wheel packages aren't being uploaded to PyPI.


codingjoe (approved):

codingjoe:
Good catch, thanks!

PR #216, opened 2019-04-15
Allow asking for JSON using a query parameter

ashokdelphia:
Thanks for this useful package. It made adding a richer health-check endpoint very straightforward.

~I'm using it in a context where a number of different agents are hitting the endpoint, and not all are easily configurable to set arbitrary headers for their requests. For us, the HTML response is about 1000 bytes, but the JSON response is about 50, so I would like to make the default response JSON instead of HTML.~

~I tried to add reasonable test coverage, and simple docs, but please let me know if you'd rather handle those differently.~

~I stopped short of fully parsing the Accept header and interpreting the q-values, as that felt like it was likely overkill for this purpose.~

This PR adds an alternative way to ask for a JSON response, by passing a `json` query parameter.

(See the comments for the change in direction here.)


codingjoe:
hi @ashokdelphia 

thanks for the contribution. You are not the first one, to approach us with that idea. Please read through #208 to see my concerns about this topic.
I am not willing to add a new setting for this feature. However I am willing to add support for a new GET attribute `?json=1`, similar to how DRF does this.

It would be great to get this one the road, just give it a little more work.

Best
-Joe

ashokdelphia:
Hi @codingjoe,

Sorry I didn't see the previous PR on a similar topic before making this.

I understand wanting to keep the number of configurable options down. I'd prefer having something that doesn't need a parameter, but it's a reasonable middle ground that will still make it possible for consumers of the endpoint where I can't set arbitrary headers.

Hopefully the latest changes are in line with what you're expecting. Please let me know if they need any adjustment.

codingjoe (approved):
@ashokdelphia I made some changes, mainly I renamed the GET attribute back to `format=json`, sorry for the back and forth.
Furthermore I change the code to actually respect the weight of the accept header media types as well as using the GET parameter as override, no matter what has been defined in the header.

Let me know if you see any problems. Otherwise, I will merge and release this feature. Thanks for all the hard work, much appreciated.

health_check/views.py
```diff
@@ -43,12 +80,18 @@ def _run(plugin):
...
+        format_override = request.GET.get('format')
 
-        context = {'plugins': plugins, 'status_code': status_code}
+        if format_override == 'json':
+            return self.render_to_response_json(plugins, status_code)
 
-        return self.render_to_response(context, status=status_code)
+        accept_header = request.META.get('HTTP_ACCEPT', '*/*')
+        for media in MediaType.parse_header(accept_header):
+            if media.mime_type in ['text/html', '	application/xhtml+xml', '*/*']:
```
ashokdelphia: It looks like there's a spurious tab in the second element here.

tests/test_views.py
```diff
@@ -4,19 +4,54 @@
...
+        assert repr(MediaType('*/*')) == "MediaType: */*;1"
+
+    def test_eq(self):
+        assert MediaType('*/*') == MediaType('*/*')
+        assert MediaType('*/*', 0.9) != MediaType('*/*')
+
+    def test_from_string(self):
+        assert MediaType.from_string('*/*') == MediaType('*/*')
+        assert MediaType.from_string('*/*;0.9') == MediaType('*/*', 0.9)
+        assert MediaType.from_string('*/*;0.9') == MediaType('*/*', 0.9)
```
ashokdelphia: I don't think the format here is following how accept headers should be formatted.

I think it should be of the form `foo/bar;q=0.9`, where the quality a parameter `q`, not a bare value.

https://tools.ietf.org/html/rfc7231#section-5.3.2

ashokdelphia: It may also be worth having explicit tests for the various places you're allowed to have optional whitespace, starting with directly after the `;`, which I think is reasonably common.

ashokdelphia (commented):
The new changes generally look good to me.

The spurious tab character is the only thing that looks clearly wonky to me.

Cheers.

tests/test_views.py
codingjoe: good lord you are right. This was wrong. I fixed it and added a bunch more tests. Please double check, before I release it.

ashokdelphia:
I adjusted the spurious whitespace, but while adding a failing test, I noticed that passing an accept header with an unrecognised type seems to cause a problem:

```
    def test_success_accept_unknown(self, client):
        class SuccessBackend(BaseHealthCheckBackend):
            def run_check(self):
                pass

        plugin_dir.reset()
        plugin_dir.register(SuccessBackend)
        response = client.get(self.url, HTTP_ACCEPT='application/octet-stream')
```

Gives:

```
===================================================== FAILURES =====================================================
_____________________________________ TestMainView.test_success_accept_unknown _____________________________________
tests/test_views.py:215: in test_success_accept_unknown
    response = client.get(self.url, HTTP_ACCEPT='application/octet-stream')
venv/lib/python3.7/site-packages/django/test/client.py:535: in get
    response = super().get(path, data=data, secure=secure, **extra)
venv/lib/python3.7/site-packages/django/test/client.py:347: in get
    **extra,
venv/lib/python3.7/site-packages/django/test/client.py:422: in generic
    return self.request(**r)
venv/lib/python3.7/site-packages/django/test/client.py:503: in request
    raise exc_value
venv/lib/python3.7/site-packages/django/core/handlers/exception.py:34: in inner
    response = get_response(request)
venv/lib/python3.7/site-packages/django/core/handlers/base.py:115: in _get_response
    response = self.process_exception_by_middleware(e, request)
venv/lib/python3.7/site-packages/django/core/handlers/base.py:113: in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
venv/lib/python3.7/site-packages/django/views/generic/base.py:71: in view
    return self.dispatch(request, *args, **kwargs)
venv/lib/python3.7/site-packages/django/views/generic/base.py:97: in dispatch
    return handler(request, *args, **kwargs)
venv/lib/python3.7/site-packages/django/views/decorators/cache.py:45: in _wrapped_view_func
    add_never_cache_headers(response)
venv/lib/python3.7/site-packages/django/utils/cache.py:252: in add_never_cache_headers
    patch_response_headers(response, cache_timeout=-1)
venv/lib/python3.7/site-packages/django/utils/cache.py:243: in patch_response_headers
    if not response.has_header('Expires'):
E   AttributeError: 'NoneType' object has no attribute 'has_header'
```

A bisect suggests it breaks in 317ad1c, but I haven't had time to dig in to why yet. I can take a look after the weekend, if it looks like a real problem and you don't have time before then.

codingjoe:
@ashokdelphia there was no fallback. I added it and a test. I also added another commit to support an empty accept header.

codingjoe:
Released in [3.10.0](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.10.0)

Thx @ashokdelphia, really good work!

PR #215, opened 2019-04-13
Test against more Django and Python versions

JayH5:
* Test against Django 2.1 & 2.2
* Install Django in tox environment from package rather than git archive
* Update default Python version in Travis to 3.7
* Test against all supported versions of Python for latest Django LTS release
* Slight Travis file cleanup (move matrix things all to one place, `jobs` is an alias for `matrix`)

This is a large-ish change to the Travis file I've made without any consultation, so I'd be totally OK if you wanted me to make fewer changes or to pick out some of these changes yourself.

PR #219, opened 2019-04-23
Clean up some old code supporting Python 2

JayH5:
Should fix tests against Django master branch.


JayH5:
`master` still failing to build but I believe it should succeed after merging #215 that updates the Travis build environment to Ubuntu Xenial with a newer SQLite.

codingjoe (approved):
Thanks @JayH5 that should also fix the builds for Django current master branch.

Issue #220, opened 2019-04-29
Accept header parsing rejects valid headers


ashokdelphia:
I'm using Chrome Version 73.0.3683.86, and the accept header it makes for regular web pages is:

```
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3
```

With the new Accept header parsing in v3.10.0, the last entry is rejected with:

```
    "application/signed-exchange;v=b3" is not a valid media type
```

Apologies for not catching that in review for PR #216.

I think to properly follow the [production in RFC 7231](https://tools.ietf.org/html/rfc7231#section-5.3.2), we should allow other parameters and only care if `q` is set.

nwolff:
Just to clarify, installing django-health-check 3.10.0 and requesting the healthcheck page returns a 500:

![2019-05-07-102102_758x251_scrot](https://user-images.githubusercontent.com/40599/57284662-eda44980-70b1-11e9-9095-7e27551e0461.png)

ashokdelphia:
> Just to clarify, installing django-health-check 3.10.0 and requesting the healthcheck page returns a 500:

Yes. #221 should fix that.

ashokdelphia:
This is fixed in v3.10.1

PR #221, opened 2019-04-29
Allow more valid accept headers

ashokdelphia:
The main aim here is to allow more valid Accept headers, particularly types with other parameters than `q`.

I also was more permissive with optional whitespace arounnd the semi-colon, and allowed up to three decimal places but only in the inclusive range from zero to one.

I haven't added any negative tests (e.g. for `q=1.500` or with too many decimal digits), just for reasons of time, but can if they'd be desirable.

See #220.


codingjoe:
@ashokdelphia this looks good. please also add the negative examples to make sure we cover all cases we can think of. Thanks again, I really appreciate the help!

health_check/views.py
```diff
@@ -18,7 +18,31 @@ class MediaType:
...
+              0([.]\d{1,3})?              # 0.000 to 0.999 with optional precision
+            )
+          )?
+          (
+            \s*;\s*                       # parameter separator with optional whitespace
+            [-!#$%&'*+.^_`|~0-9a-zA-Z]+   # any token from legal characters
+            =
+            [-!#$%&'*+.^_`|~0-9a-zA-Z]+   # any value from legal characters
+          )*
+          $
```
ashokdelphia, line 44: I'm tempted to replace lines 38–44 with just:

```
          (\s*;|$)                        # followed by other parameters or the end of string                                                   
```

And not care about the structure of the other parameters. This would cause the `*/*;0.9` test to fail, but I wonder if that's preferable.

ashokdelphia:
> @ashokdelphia this looks good. please also add the negative examples to make sure we cover all cases we can think of. 

Only in trying to add cases for invalid q-values, I noticed that they're still permitted by the last clause in the regular expression. I'm not sure it's worth making that more fierce in order to reject them.

I did add a few other failing cases, though, and left a structure where dropping more in is reasonably straightforward. 

> Thanks again, I really appreciate the help!

You're very welcome. Cheers.

health_check/views.py
codingjoe: Your call, but I can live with complex regular expressions.

codingjoe (approved):
LGTM, thanks @ashokdelphia

codingjoe:
@ashokdelphia let me know how you decide on the open question you had. I will merge and release once you decided.

ashokdelphia:
> @ashokdelphia let me know how you decide on the open question you had. I will merge and release once you decided.

Let's go with what we have, then. If we find we hit any other error cases, we can make it looser about the other parameters later.

codingjoe:
Released in [3.10.1](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.10.1)

PR #222, opened 2019-05-15
Improve accept header handling for unknown types

ashokdelphia:
In the project where I'm using this, I particularly care about the case where the request prefers other types but has a lower quality match for JSON, e.g.:

    Accept: application/x-unknown, application/json; q=0.1

At the moment, it resolves using the first entry in the sorted list of types, even if there are better matches later on.

Sorry I failed to spot this when reviewing the q-value changes for #216.

### DIFF
--- a/health_check/views.py
+++ b/health_check/views.py
@@ -126,8 +126,7 @@ class MainView(TemplateView):
                 return self.render_to_response(context, status=status_code)
             elif media.mime_type in ('application/json', 'application/*'):
                 return self.render_to_response_json(plugins, status_code)
-            else:
-                return self.render_to_response(context, status=status_code)
+        return self.render_to_response(context, status=status_code)
 
     def render_to_response_json(self, plugins, status):
         return JsonResponse(
--- a/health_check/views.py
+++ b/health_check/views.py
@@ -2,8 +2,7 @@ import copy
 import re
 from concurrent.futures import ThreadPoolExecutor
 
-from django.http import JsonResponse
-from django.http import HttpResponse
+from django.http import HttpResponse, JsonResponse
 from django.views.decorators.cache import never_cache
 from django.views.generic import TemplateView
 
--- a/health_check/views.py
+++ b/health_check/views.py
@@ -3,6 +3,7 @@ import re
 from concurrent.futures import ThreadPoolExecutor
 
 from django.http import JsonResponse
+from django.http import HttpResponse
 from django.views.decorators.cache import never_cache
 from django.views.generic import TemplateView
 
@@ -126,7 +127,7 @@ class MainView(TemplateView):
                 return self.render_to_response(context, status=status_code)
             elif media.mime_type in ('application/json', 'application/*'):
                 return self.render_to_response_json(plugins, status_code)
-        return self.render_to_response(context, status=status_code)
+        return HttpResponse('Only HTML and JSON responses are supported', status=406, content_type='text/plain')
 
     def render_to_response_json(self, plugins, status):
         return JsonResponse(
--- a/health_check/views.py
+++ b/health_check/views.py
@@ -126,7 +126,7 @@ class MainView(TemplateView):
                 return self.render_to_response(context, status=status_code)
             elif media.mime_type in ('application/json', 'application/*'):
                 return self.render_to_response_json(plugins, status_code)
-        return HttpResponse('Only HTML and JSON responses are supported', status=406, content_type='text/plain')
+        return HttpResponse('Not Acceptable: Supported content types: text/html, application/json', status=406, content_type='text/plain')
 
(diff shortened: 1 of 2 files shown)


health_check/views.py
```diff
@@ -126,8 +126,7 @@ def _run(plugin):
                 return self.render_to_response(context, status=status_code)
             elif media.mime_type in ('application/json', 'application/*'):
                 return self.render_to_response_json(plugins, status_code)
-            else:
-                return self.render_to_response(context, status=status_code)
+        return self.render_to_response(context, status=status_code)
```
codingjoe: How can that ever happen if the default is `*/*` see a couple lines above?
The only way this would happen, is if someone sends an empty accept header. In this case it might probably sense to respond with a 400.

tests/test_views.py
```diff
@@ -139,6 +139,16 @@ def run_check(self):
         response = client.get(self.url, HTTP_ACCEPT='application/json')
         assert response['content-type'] == 'application/json'
 
+    def test_success_prefer_json(self, client):
```
codingjoe, line 143: The more the merrier, but I believe we have that already covered, don't we?

tests/test_views.py
```diff
@@ -159,6 +169,16 @@ def run_check(self):
         response = client.get(self.url, HTTP_ACCEPT='application/octet-stream')
         assert response['content-type'] == 'text/html; charset=utf-8'
 
+    def test_success_unsupported_and_supported_accept(self, client):
```
codingjoe, line 177: Same here, I believe this is covered, isn't it?

codingjoe (commented):
Hi @ashokdelphia good to see that you are still following up on this issue. I did leave a couple of comments. Let me know what you think.

health_check/views.py
ashokdelphia: The Accept header can be absent, or could just contain types that we don't care about in the conditions above.

There's no requirement, as far as I remember, that the caller includes `*/*` in their list, so I think it's legitimate to fall off the bottom of the loop.

tests/test_views.py
ashokdelphia: This test was failing with the old code, because the loop was only considering the first element in the sorted list.

tests/test_views.py
ashokdelphia: It is, but I don't think this particular case is quite there in the other tests.

health_check/views.py
codingjoe: I would honestly like to see a test that breaks it and I still believe in this case we should respond with a 400, since we can't serve a response that would satisfy the request.

ashokdelphia: Apologies. I thought I replied to this yesterday.

The old code did return HTML if the 'top' result from the request's accept list didn't match the explicit lists, so my change here was trying to keep that behaviour, but keep looking at the rest of the acceptable types.

I think you're right, though: it's reasonable to make that case a 400-series response. I think a 406 would be appropriate. I'll add a covering test and change the behaviour.

codingjoe: @ashokdelphia, OK, go for it!

ashokdelphia: Sorry for the delay responding.

I pushed up the changes.

I'm a little concerned this would be a breaking change for anyone currently hitting the endpoint with an unrecognized Accept header, but if you think that's okay, then feel free to merge.

health_check/views.py
```diff
@@ -126,8 +126,7 @@ def _run(plugin):
                 return self.render_to_response(context, status=status_code)
             elif media.mime_type in ('application/json', 'application/*'):
                 return self.render_to_response_json(plugins, status_code)
-            else:
-                return self.render_to_response(context, status=status_code)
+        return HttpResponse('Only HTML and JSON responses are supported', status=406, content_type='text/plain')
```
codingjoe:
```suggestion
        return HttpResponse('Not Acceptable: Supported content types: plain/html, application/json', status=http.HTTPStatus.NOT_ACCEPTABLE, content_type='text/plain')
```

codingjoe (approved):
Cool, feature wise, this looks perfect. I just made some suggestions regarding the text.
Let me know once you are happy and I will merge and release it.

health_check/views.py
ashokdelphia: Thanks for the suggestion. I applied it with a minor tweak.

ashokdelphia:
@codingjoe I think that's good to go now, if you're happy with it.

codingjoe:
@ashokdelphia I'll merge this once you fix the test :)

ashokdelphia:
Apologies for missing the lint failure. That should be fine now.

codingjoe:
Released in [3.10.2](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.10.2)

PR #225, opened 2019-06-27
Health Check for Redis

ingvaldlorentzen:
This adds support for a health check for Redis. 
More or less the same as the health check for RabbitMQ. 

It currently uses the `celery_broker_url` from Django settings. This could be changed to `broker_url` that RabbitMQ uses, if wanted.

### DIFF
--- a/health_check/contrib/redis/backends.py
+++ b/health_check/contrib/redis/backends.py
@@ -12,10 +12,7 @@ logger = logging.getLogger(__name__)
 class RedisHealthCheck(BaseHealthCheckBackend):
     """Health check for Redis."""
 
-    def __init__(self):
-        super(RedisHealthCheck, self).__init__()
-        logger.debug("Checking for a redis_url in django settings...")
-        self.redis_url = getattr(settings, "REDIS_URL", None)
+    redis_url = getattr(settings, "REDIS_URL", 'redis://localhost/1')
 
     def check_status(self):
         """Check Redis service by pinging the redis instance with a redis connection."""
--- a/health_check/contrib/redis/backends.py
+++ b/health_check/contrib/redis/backends.py
@@ -24,10 +24,7 @@ class RedisHealthCheck(BaseHealthCheckBackend):
         try:
             # conn is used as a context to release opened resources later
             with from_url(redis_url) as conn:
-                print('conn', conn)
                 conn.ping()  # exceptions may be raised upon ping
-                print(conn.ping())
-                print('got here')
         except ConnectionRefusedError as e:
             self.add_error(ServiceUnavailable("Unable to connect to Redis: Connection was refused."), e)
         except exceptions.TimeoutError as e:
--- a/health_check/contrib/redis/backends.py
+++ b/health_check/contrib/redis/backends.py
@@ -12,8 +12,10 @@ logger = logging.getLogger(__name__)
 class RedisHealthCheck(BaseHealthCheckBackend):
     """Health check for Redis."""
 
-    logger.debug("Checking for a redis_url in django settings...")
-    redis_url = getattr(settings, "REDIS_URL", None)
+    def __init__(self):
+        super(RedisHealthCheck, self).__init__()
+        logger.debug("Checking for a redis_url in django settings...")
+        self.redis_url = getattr(settings, "REDIS_URL", None)
 
     def check_status(self):
         """Check Redis service by pinging the redis instance with a redis connection"""
--- a/health_check/contrib/redis/backends.py
+++ b/health_check/contrib/redis/backends.py
@@ -1,7 +1,7 @@
 import logging
 
 from django.conf import settings
-from redis import from_url, exceptions
+from redis import exceptions, from_url
 
 from health_check.backends import BaseHealthCheckBackend
 from health_check.exceptions import ServiceUnavailable
--- a/health_check/contrib/redis/backends.py
+++ b/health_check/contrib/redis/backends.py
@@ -1,7 +1,7 @@
 import logging
 
 from django.conf import settings
-from redis import from_url, exceptions
+from redis import exceptions, from_url
 
 from health_check.backends import BaseHealthCheckBackend
 from health_check.exceptions import ServiceUnavailable
@@ -24,7 +24,10 @@ class RedisHealthCheck(BaseHealthCheckBackend):
         try:
             # conn is used as a context to release opened resources later
             with from_url(redis_url) as conn:
+                print('conn', conn)
                 conn.ping()  # exceptions may be raised upon ping
+                print(conn.ping())
+                print('got here')
         except ConnectionRefusedError as e:
             self.add_error(ServiceUnavailable("Unable to connect to Redis: Connection was refused."), e)
         except exceptions.TimeoutError as e:
--- a/health_check/contrib/redis/backends.py
+++ b/health_check/contrib/redis/backends.py
@@ -2,6 +2,7 @@ import logging
 
 from django.conf import settings
 from kombu import Connection
+from kombu.exceptions import TimeoutError, ConnectionLimitExceeded
 
 from health_check.backends import BaseHealthCheckBackend
 from health_check.exceptions import ServiceUnavailable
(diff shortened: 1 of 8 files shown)


codingjoe:
@ingvaldlorentzen thanks for the contribution. I think a Redis health-check would be great. That being said I see multiple places where Redis can be used in Django. I for one, use it for caching, not as a result backend for celery. So either we change implementation, or we change the naming a little to reflect, that it's monitoring the result backend.

Please also drop the commented out lines ;)

Thanks!
Joe

ingvaldlorentzen:
Hi, 

Thanks for the feedback!
I changed the Redis URL from `CELERY_BROKER_URL` to simply `REDIS_URL`. Also updated the variables in the health check to reflect this change. 
I think `REDIS_URL ` is a nice generic way to do this, and if a project doesn't use `REDIS_URL`, it's easy and hopefully understandable to add to the settings variables for the health check. 
Removed the commented code, sorry about that, it was a slip-up.

I'll add info on the Redis health check to the documentation if you think this looks alright. 

Regards, 
Ingvald

**Edit**: I see some checks are failing here. I'm not super familiar with codecov, if you have any feedback on how I should proceed to fix this I would be super thankful!
If not, I'll see if I can get a team member to help me out on Monday.

health_check/contrib/redis/backends.py
```diff
@@ -0,0 +1,37 @@
+import logging
+
+from django.conf import settings
+from kombu import Connection
```
codingjoe: `kombu` is part of celery and current not listed in the requirements. So maybe it makes sense to switch to another redis client. There also should be documentation, or an `extra` requirement addtion to the setuptools config.

health_check/contrib/redis/backends.py
```diff
@@ -0,0 +1,37 @@
...
+
+        logger.debug("Attempting to connect to redis...")
+        try:
+            # conn is used as a context to release opened resources later
+            with Connection(redis_url) as conn:
+                conn.connect()  # exceptions may be raised upon calling connect
+        except ConnectionRefusedError as e:
+            self.add_error(ServiceUnavailable("Unable to connect to Redis: Connection was refused."), e)
+
+        except IOError as e:
```
codingjoe: You should be more verbose about what error you are expecting here.

health_check/contrib/redis/backends.py
```diff
@@ -0,0 +1,37 @@
...
+            self.add_error(ServiceUnavailable("IOError"), e)
+
+        except BaseException as e:
```
codingjoe, line 32: Please don't catch base exceptions, those are already handled by the `BaseHealthCheckBackend`

tests/test_redis.py
```diff
@@ -0,0 +1,30 @@
+import mock
+
+from health_check.contrib.redis.backends import RedisHealthCheck
+
+
+class TestRabbitMQHealthCheck:
```
codingjoe:
```suggestion
class TestRedisHealthCheck:
```
;)

tests/test_redis.py
```diff
@@ -0,0 +1,30 @@
...
+    @mock.patch("health_check.contrib.redis.backends.getattr")
+    @mock.patch("health_check.contrib.redis.backends.Connection")
+    def test_redis_refused_connection(self, mocked_connection, mocked_getattr):
+        """Test when the connection to Redis is refused."""
+        mocked_getattr.return_value = "redis_url"
+
+        conn_exception = ConnectionRefusedError("Refused connection")
+
+        # mock returns
+        mocked_conn = mock.MagicMock()
```
codingjoe: You test only one exception type here. Please test all of them.

codingjoe (changes requested):
@ingvaldlorentzen I like the changes you made. There certainly needs to be some documentation added for the setting and health check as a whole. I think this is on a good path :) Keep up the good work!

health_check/contrib/redis/backends.py
JonasKs: I removed it entirely, I think the main reason for it to be there was due to the `RabbitMQ` backend had it. Not sure why it would be Excepted specifically. Please let me know if you want it added back.

health_check/contrib/redis/backends.py
JonasKs: Fixed.

tests/test_redis.py
JonasKs: Fixed.

tests/test_redis.py
JonasKs: Fixed. 
I see `patch` and `project` still fails. I can't see `rabbitmq.apps` being tested, so I'm not sure if you want this to be tested too?

health_check/contrib/redis/backends.py
```diff
@@ -0,0 +1,36 @@
...
+            with Connection(redis_url) as conn:
+                conn.connect()  # exceptions may be raised upon calling connect
+        except ConnectionRefusedError as e:
+            self.add_error(ServiceUnavailable("Unable to connect to Redis: Connection was refused."), e)
+        except exceptions.TimeoutError as e:
+            self.add_error(ServiceUnavailable("Unable to connect to Redis: Timeout."), e)
+        except exceptions.ConnectionLimitExceeded as e:
+            self.add_error(ServiceUnavailable("Unable to connect to Redis: "
+                                              "Maximum number of simultaneous connections exceeded"), e)
+        else:
```
JonasKs, line 34: Hard to have codecov on this, since there's nothing to be returned or anything.

health_check/contrib/redis/backends.py
```diff
@@ -0,0 +1,36 @@
+import logging
+
+from django.conf import settings
+from kombu import Connection, exceptions
```
codingjoe: `kombu` is still not listed as a requirement. I also don't think that you need `kombu` since it's a messaging library. I would prefer to use plain [redis](https://pypi.org/project/redis/).

health_check/contrib/redis/backends.py
codingjoe: Just write a test case, where nothing fails. You can use py.test's [caplog](https://docs.pytest.org/en/latest/logging.html) to make assertions on the debug log.

health_check/contrib/redis/backends.py
```diff
@@ -0,0 +1,36 @@
...
+                conn.connect()  # exceptions may be raised upon calling connect
+        except ConnectionRefusedError as e:
+            self.add_error(ServiceUnavailable("Unable to connect to Redis: Connection was refused."), e)
+        except exceptions.TimeoutError as e:
+            self.add_error(ServiceUnavailable("Unable to connect to Redis: Timeout."), e)
+        except exceptions.ConnectionLimitExceeded as e:
+            self.add_error(ServiceUnavailable("Unable to connect to Redis: "
+                                              "Maximum number of simultaneous connections exceeded"), e)
+        else:
+            logger.debug("Connection established. Redis is healthy.")
```
codingjoe, line 35: I would use `info` over `debug` here.

codingjoe (changes requested):
Almost there! Keep up the good work @ingvaldlorentzen

health_check/contrib/redis/backends.py
ingvaldlorentzen: Redis library implemented instead of kombu

health_check/contrib/redis/backends.py
ingvaldlorentzen: Added some documentation to the README, not a whole lot, but not sure if any more is really necessary.

JonasKs:
I tried to create the tests for most things now that @ingvaldlorentzen has refactored over to use `redis`.

I haven't used pytest before, so please give me a heads up if I need to correct something.  
`test_redis_conn_ok` could probably be rewritten to catch the log message as you requested, but I'd have to look more into that in two weeks when I'm back from vacation.

health_check/contrib/redis/backends.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class RedisHealthCheck(BaseHealthCheckBackend):
+    """Health check for Redis."""
+
+    def check_status(self):
+        """Check Redis service by opening and closing a connection."""
+        logger.debug("Checking for a redis_url in django settings...")
+
+        redis_url = getattr(settings, "REDIS_URL", None)
```
codingjoe: Can you make that a class attribute? I for one, have multiple Redis servers running, so I will need to subclass this health check.

health_check/contrib/redis/backends.py
```diff
@@ -0,0 +1,37 @@
...
+        except ConnectionRefusedError as e:
+            self.add_error(ServiceUnavailable("Unable to connect to Redis: Connection was refused."), e)
+        except exceptions.TimeoutError as e:
+            self.add_error(ServiceUnavailable("Unable to connect to Redis: Timeout."), e)
+        except exceptions.ConnectionError as e:
+            self.add_error(ServiceUnavailable("Unable to connect to Redis: Connection Error"), e)
+        except BaseException as e:
+            self.add_error(ServiceUnavailable("Unknown error"), e)
+        else:
+            logger.info("Connection established. Redis is healthy.")
```
codingjoe:
```suggestion
            logger.debug("Connection established. Redis is healthy.")
```

codingjoe (changes requested):
Looks very good,  I just had some final comments.

health_check/contrib/redis/backends.py
ingvaldlorentzen: Good idea! Added it

health_check/contrib/redis/backends.py
ingvaldlorentzen: Fixed.

codingjoe:
Released in [3.11.0](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.11.0) 🎉

PR #230, opened 2019-09-17
Added support for redis 2.10.6 check

rohit-innovaccer:
Currently, the code breaks for redis 2.10.6, but we have to use that version of redis only since its the one supported by celery. So I just changed the way to check the status of redis.
Also, for celery check, there is a case where we don't have the celery queues defined in celery app, in that case, we cannot check the status of those queues, so just added a way to check the status of queues using a settings variable which contains the list of queues used in your project.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -110,12 +110,10 @@ on django.conf.settings with the required format to connect to your redis server
 To use the Celery healthcheck, if checking just for default celery queue, no need to do any changes.
 
 But if you want to check for some more queues along with default ones, please make sure that there
-are 2 variables named ``USE_DEFAULT_QUEUE`` and ``CELERY_QUEUES`` on django.conf.settings and set the value of ``USE_DEFAULT_QUEUE`` to False and ``CELERY_QUEUES`` to ['A', 'B'] where 'A' and 'B' are celery queues.
-please make sure that there is a variable named ``REDIS_HOST``
+is a variables named ``CELERY_QUEUES`` on django.conf.settings and set its value to ['A', 'B'] where 'A' and 'B' are celery queues.
 
 .. code::
 
-    USE_DEFAULT_QUEUE = False
     CELERY_QUEUES = ['A', 'B']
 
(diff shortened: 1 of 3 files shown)


codingjoe:
@rohit-innovaccer thanks for reaching out. You help really is much appreciated. I just wish you would have opened an issue before going through all the trouble of writing a patch.

After a brief review, I have two main reservations:
1. You can pass the Redis port as part of the URL, like so: `redis://localhost:MYPORT/1`.
2. I fail to see, how the change on the celery backend is related. More importantly, multi queue support was introduced in version 3.2.0, see also: https://github.com/KristianOellegaard/django-health-check/commit/8a86a1bc80bf6f38549da82a0e3875ec8acf69eb

Right now, I see no problem that isn't already solved. Therefore, I will close the PR. Please let me know if you have any further questions, or if I missed anything.

Best
-Joe

codingjoe:
@rohit-innovaccer little side note, since you seemed to have trouble with both of those features, we might have a documentation issue. I would really appreciate if you could find the time to improve our documentation.

rohit-innovaccer:
@codingjoe Hi, Along with giving redis post, I also changed the function currently being used for redis healthcheck, since it was breaking for redis 2.10.6.
Also, regarding celery, there is a case where I had not defined celery queues in my code that i'll be going to use, so in app.amqp.queues I do not get my queues, for that case, I cannot check my celery queues.
And I raised PR directly because I took a fork of your library and made changes according to my usecase, then I thought others might face the same issue so i raised the PR.

codingjoe:
Hi @rohit-innovaccer,

thanks for the swift response. Let me try to unpack that a bit, to make sure I understood you correctly. Your patch fixes an issue that only exists in a very old Redis version, right? What would that issue be exactly?
Also `app.amqp.queues` doesn't seem to work for you, why is that? Can you maybe share you Celery setup here, so I can understand the problem better?

Thanks
-Joe

PR #235, opened 2020-01-14
Fix #233 -- Include original exception data

### LINKED ISSUE #233: Catching exception hides true cause
Hi,

There is a [`try` block](https://github.com/KristianOellegaard/django-health-check/blob/master/health_check/storage/backends.py#L69) that seems to mask the true cause of an exception in the backend, which makes it hard to diagnose why the health check fails. I believe that by allowing the normal error to bubble up or by including the original exception in the `ServiceUnavailable` exception, it would be easier for dev/devops teams to diagnose why a service is down.

codingjoe:


lizzielizzielizzie:
This looks perfect! Thank you

codingjoe:
Released in [3.12.1](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.12.1) 🎉

PR #237, opened 2020-01-17
Fix isort

toopy:
Unfortunately I think I introduced that ones. Sorry about that.


codingjoe:
Thanks @toopy. Apparently some checks don't run anymore. @KristianOellegaard can you make sure you accepted to updated permissions for [FussyFox](https://github.com/apps/fussy-fox)?

PR #243, opened 2020-03-12
Fix typo in health_check command documentation

mastacheata:
Causing a command to quit has nothing to do with capitalization, boxes or conditionals, so it should be written with a u between a and s ;)


codingjoe (approved):
Good catch! Thanks @mastacheata

PR #244, opened 2020-03-16
non default db connection bug patch

ashexpertVersion2:


health_check/mixins.py
```diff
@@ -33,8 +33,9 @@ def _run(plugin):
             try:
                 return plugin
             finally:
-                from django.db import connection
-                connection.close()
+                from django.db import connections
+                for connection in connections.all():
+                    connection.close()
```
codingjoe: There is actually a method on the `ConnectionHandler` that is very similar, but also add some exception handling for already closed connections.
```suggestion
                from django.db import connections
                connections.close_all()
```

tests/testapp/__init__.py
```diff
@@ -1,3 +1,3 @@
-from .celery import app
+from .celery_app import app
```
codingjoe: This change is not related and also not needed. Please revert the change.
```suggestion
from .celery import app
```
You will also need to rename the file again.

tests/testapp/settings.py
```diff
@@ -9,6 +9,10 @@
     'default': {
         'ENGINE': 'django.db.backends.sqlite3',
         'NAME': ':memory:',
+    },
+    'backup': {
```
codingjoe: IMHO, we might want to give it a more generic name and a comment, to not confuse people.
```suggestion
    'other': {  # 2nd database conneciton to ensure proper connection handling
```

codingjoe (commented):
Hi @ashexpertVersion2 thanks for the contribution. You are off to a very good start. I did add some comments that might help to improve your patch. Please let me know if you have any questions.
Best -Joe

tests/testapp/__init__.py
ashexpertVersion2: i couldn't run tests due to circular dependency caused by similarity of this file's name and the celery module. you are right i shouldn't add it to this commit but it slipped out of my hand.

codingjoe (approved):
LGTM Thanks again @ashexpertVersion2 much appreciated! 👍

codingjoe:
Released in [3.12.2](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.12.2) 🎉

ashexpertVersion2:
Your welcome :+1: 
thank you for fast response.

PR #245, opened 2020-04-27
make example simplifier to copy and paste

ehmkah:
Small changes to example, make it easier to copy and paste it.


codingjoe:
Good catch @ehmkah, we should probably enable doctests to make sure those errors don't appear again. Anyways, thanks a lot! Good documentation is crucial.

PR #251, opened 2020-05-24
Improve GitHub Actions PyPI release trigger (refs #250)

amureki:
Hey @codingjoe!

Thanks for actively maintaining this and plenty of other packages, you are the champion! :)
I was replacing Travis CI with GitHub actions in one of my packages (https://github.com/amureki/statuscheck) and was inspired by your code here in #250.

However, I noticed a couple of things that were not correct from my perspective:

- release job was triggered three times (fixed by [specifying exact action type "published"](https://help.github.com/en/actions/reference/events-that-trigger-workflows#release-event-release))
~~- TWINE_USERNAME variable was coming from `__token__` - not sure I follow, how this works, but I changed it to `secrets.TWINE_USERNAME`~~
- (other) I put a couple of naming suggestions, feel free to disregard

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,20 +1,22 @@
-name: Release
+name: PyPI release
 
-on: [release]
+on:
+  release:
+    types: [released]
 
 jobs:
-
-  PyPi:
+  PyPI:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
-    - name: Install Python dependencies
-      run: python -m pip install --upgrade pip setuptools wheel twine
-    - name: Build dist packages
-      run: python setup.py sdist bdist_wheel
-    - name: Upload packages
-      run: python -m twine upload dist/*
-      env:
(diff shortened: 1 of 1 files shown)


codingjoe:
Hi @amureki how are you doing? Let me try to address your questions, before I get into the code:

1. The release types were added recently it seems. I believe the proper type would be `published` though, instead of `released`. 
2. `__token__` is the correct username if you use the new [PyPi tokens](http://pyfound.blogspot.com/2019/07/pypi-now-supports-uploading-via-api.html).
3. I would not want to rename the workflow itself. Release is a good name and `PyPi` a good name for the job. I have other repos I maintain like select2 where I have to release to different registries, so there I have an npm job as well.

So currently the release didn't work. I can't change the repos secrets only @KristianOellegaard can do so. We talked via email, it should be set, but something went wrong. Kind of hard to debug. Having encrypted variables in code, like on Travis, sure had its advantages.

Best,
Johannes

codingjoe (approved):
LGTM, thanks!

amureki:
@codingjoe thanks for clarification about tokens and for the merge!

PR #252, opened 2020-06-02
Added backend for checking if all migrations are applied

su-tiko:
This PR adds a new Health Check by using `MigrationExecutor` for checking if there are pending migrations to apply.

I think that this HealthCheck combined with the `db`  cover a wide range of scenarios.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,6 +16,7 @@ The following health checks are bundled with this project:
 - AWS S3 storage
 - Celery task queue
 - RabbitMQ
+- Migrations
 
 Writing your own custom health checks is also very quick and easy.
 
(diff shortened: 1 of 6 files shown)


codingjoe:
Interesting, can you elaborate a bit more on the use case? For me this is usually part of the deployment routine and my apps do not deploy with the migrations passing. So I am curious from whom this is intended.

su-tiko:
I'm using it on kubernetes service so old pods are not destroyed before new migrations are applied. Once applied, new pods get the `ready` state and become coming up.

So my deployment pipeline runs as follow:

-> Launch new pods -> Run migrations job -> Destroy old pods

codingjoe:
I see, well the code looks spotless, released in [3.13.3](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.13.3)

I messed up the version number, but it's a new minor release, so we should be good.

su-tiko:
Hi! Do you plan to upgrade PyPi version soon? Just for switching my requirements back to raw package name.

Thx!

codingjoe:
there is nothing I can do right now. I am waiting on feedback from @KristianOellegaard

PR #253, opened 2020-06-20
Update Readme instructions with boto3 version instead of boto2

anushbmx:
Update Readme with boto3 version instead of boto2 as boto2 is deprecated.


codingjoe (approved):
Good catch @anushbmx thanks! 👍

PR #255, opened 2020-07-21
Mongo health check backend

soul4code:


codingjoe:
@magnitronus thanks for the contribution, much appreciated. Would you mind fixing the test suite before assigning this for review again. Oh, and maybe add a like to the README about the new health check. Thanks!

PR #257, opened 2020-07-29
Clean results task Health Check

frankyjquintero:
It is necessary a elimination of the task after its verification to avoid the accumulation of results in views such as flower or saturating the backend


frankyjquintero:
@auvipy

codingjoe:
Hi @frankyjquintero thanks for the contribution. Good catch. Sadly the release is broken right now, I hope to make this available to the public asap.

PR #258, opened 2020-08-02
Fix isort

frankyjquintero:
Fix isort


frankyjquintero:
@codingjoe please review new pull request

PR #262, opened 2020-08-12
Added more detail for celery task timeouts

lizzielizzielizzie:
Hello. I created this pull request because we were having difficulties where the celery health check backend was failing but it was challenging to inspect why. To address this, I have added:

- More detailed message for cases where a task takes too long to start (waits too long in queue)
- More detailed message for cases where a task takes too long to return a result (waits to long in queue or takes too long to recieve result... basically, the original timeout behavior)
- Added settings. Both default to the value of the original HEALTHCHECK_CELERY_TIMEOUT (which defaults to 3), because not all use cases will want split control, and to maintain backwards compatibility. 
  - HEALTHCHECK_CELERY_QUEUE_TIMEOUT, which sets the maximum time for a task to wait in queue
  - HEALTHCHECK_CELERY_RESULT_TIMEOUT, which sets the maximum time for a task to return a result (queue + execution + response time)

I decided to add the two new settings because my particular celery backend doesn't report tasks that expire as failures and eventually discards them entirely, which makes it really challenging to identify *why* the tasks are expiring in the queue. With the finer control, we'll be able to extend the max queue time without impacting the acceptable performance criteria (i.e., while making sure health checks still fail if jobs take too long)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -287,32 +287,6 @@ This should yield the following output:
 
 Similar to the http version, a critical error will cause the command to quit with the exit code `1`.
 
-
-Advanced Configuration
-----------------------
-Using `django.settings` you may exert more fine-grained control over the behavior of `django-health-check`.
-
-.. list-table:: Additional Settings
-   :widths: 25 10 10 55
-   :header-rows: 1
-
-   * - Name
-     - Type
-     - Default
-     - Description
-   * - `HEALTHCHECK_CELERY_TIMEOUT`
-     - Number
-     - 3
-     - Base timeout, in seconds, for a celery task to complete. Depricated.
-   * - `HEALTHCHECK_CELERY_QUEUE_TIMEOUT`
-     - Number
-     - `HEALTHCHECK_CELERY_TIMEOUT` (3)
-     - Overrides `HEALTHCHECK_CELERY_TIMEOUT`, specifying the maximum amount of time a task may spend in the queue before being automatically revoked with a `TaskRevokedError`.
-   * - `HEALTHCHECK_CELERY_RESULT_TIMEOUT`
-     - Number
-     - `HEALTHCHECK_CELERY_TIMEOUT` (3)
-     - Overrides `HEALTHCHECK_CELERY_TIMEOUT`, specifying the maximum total time for a task to complete and return a result.
-
 Other resources
 ---------------
 
--- a/README.rst
+++ b/README.rst
@@ -287,6 +287,7 @@ This should yield the following output:
 
 Similar to the http version, a critical error will cause the command to quit with the exit code `1`.
 
+
 Other resources
 ---------------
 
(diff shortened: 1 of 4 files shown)


codingjoe (commented):
Hi @NickWer, thanks for the contribution. It certainly makes sense to split those two timeout settings. Could you please add them to the documentation as well? Since the previous setting was not documented, I think we can simply remove it. If you think it's better to keep it, you should at least add a deprecation warning. I don't think this package needs deprecation though, a major release should be warning enough. Best, Joe

lizzielizzielizzie:
So I believe this is done, but if you'd like tomorrow I can probably spend some time addressing the code coverage if you'd like

health_check/contrib/celery/apps.py
```diff
@@ -9,6 +12,11 @@ class HealthCheckConfig(AppConfig):
 
     def ready(self):
         from .backends import CeleryHealthCheck
+        if hasattr(settings, "HEALTHCHECK_CELERY_TIMEOUT"):
```
codingjoe, line 15: I'd rather use `warnings.warn("foo bar", DeprecationWarning)` here over a warning log.

README.rst
```diff
@@ -288,6 +288,31 @@ This should yield the following output:
 Similar to the http version, a critical error will cause the command to quit with the exit code `1`.
 
 
+Advanced Configuration
```
codingjoe: We have a `docs/settings.rst`. I feel this should rather go there, instead of the readme.

codingjoe (changes requested):
Great, almost there. Just two tiny things and I believe we are good.

health_check/contrib/celery/apps.py
lizzielizzielizzie: Hadn't heard of this, very neat!

lizzielizzielizzie:
Hi, sorry it took so long to get back to this - life happened and then I forgot about it 😅 anyways, I hope the changes are more satisfactory this time around, but let me know if not and I'll do my best to correct things sooner lmao

docs/settings.rst
```diff
@@ -69,3 +69,28 @@ exceeds 90% or available memory drops below 100 MB.
...
+
+.. list-table:: Additional Settings
+   :widths: 25 10 10 55
+   :header-rows: 1
+
+   * - Name
+     - Type
+     - Default
+     - Description
+   * - `HEALTHCHECK_CELERY_TIMEOUT`
```
codingjoe: You don't need to document `HEALTHCHECK_CELERY_TIMEOUT` since it deprecated. Only document the behavior for the new settings. You can also ignore the fallback in the documentation.

codingjoe (commented):
This looks great, almost there. I have only one remaining comment. Thanks!

docs/settings.rst
lizzielizzielizzie: 🎉 updated

codingjoe (approved):
LGTM, thanks @NickWer for your effort 👍

codingjoe:
Released in [3.15.0](https://pypi.org/project/django-health-check/3.15.0/) 🎉

PR #264, opened 2020-08-15
replace travis CI with github workflow

skarzi:
In this PR I have replaced travis CI pipeline with GitHub workflow as discussed in #260 

What do you think about removing tox? 
We can rely on jobs matrix and some `Makefile` or similar solution with stages for most common use cases etc.
I think that these changes can make contributing to the projects easier.


codingjoe (commented):
Hi @skarzi thanks a lot for the effort. I took a first glance and these are my points to discuss.

1. I'd like to ditch tox and rather use proper GitHub action jobs with a matrix.
2. I usually prefer the codecov python page over the codecov action for Python builds. It's a lot faster and both are officially supported.
3. You can ditch the badges for all I care. They are somewhat useless, since commit status are always shown on GitHub.

skarzi:
I will apply all your suggestions in next commits, later this week :+1:

codingjoe:
Hi @skarzi thanks for the contribution. I added some small changes and squashed your commits. Best, Joe

PR #276, opened 2020-10-21
Decode redis_url

PMExtra:
For now, it won't work correctly with any redis_url that contains uri reserved characters (usually in password).

According to [RFC3986 Section 3.2.1](https://tools.ietf.org/html/rfc3986#section-3.2.1):
> userinfo    = *( unreserved / pct-encoded / sub-delims / ":" )

So any reserved characters in userinfo must be encoded. The same, any '%' in uri should be regard as encoded-escape character.

There is a `reids-py@4.0` version that still in development has fixed it, `redis-py@4.0` will always decode url. https://github.com/andymccurdy/redis-py/issues/589

But in the stable `redis-py`, it only provides an optional argument `decode_components=False` to implement url decoding. So we should set it `True`.

Another accordance: [celery](https://github.com/celery/celery) does decode redis_url automatically.

### DIFF
--- a/health_check/contrib/redis/backends.py
+++ b/health_check/contrib/redis/backends.py
@@ -21,7 +21,7 @@ class RedisHealthCheck(BaseHealthCheckBackend):
         logger.debug("Attempting to connect to redis...")
         try:
             # conn is used as a context to release opened resources later
-            with from_url(self.redis_url) as conn:
+            with from_url(self.redis_url, decode_components=True) as conn:
                 conn.ping()  # exceptions may be raised upon ping
         except ConnectionRefusedError as e:
             self.add_error(ServiceUnavailable("Unable to connect to Redis: Connection was refused."), e)


PMExtra:
It works fine with `Django==2.2.12` and `redis==3.3.11`, But CI checks failed because of mock.

codingjoe:
Hi @PMExtra, this seems to be an issue in redis not in this package. The fact that the default will change in version 4.0 is a good indicator. I don't know how I feel about adding a backport here, since redis considers this a backwards incompatible change, it would be backwards incompatible here too. What do you think? Best, Joe

PMExtra:
I admit it's a backward incompatible change, but it only affect the project which used '%' in connection url and assume it's raw '%' character instead of escape keyword. I think this situation is almost non-existent.
RFC has stipulated any reserved character in URI should be pct-encoded. And most of libraries such as celery will decode it.

To say the least, if this pr merged, the broken project can be fix by just encode the connection url (just replace '%' with '%25'). But the projects that connected to redis with special symbols in the password cannot work with django-health-check without this PR.

For example, my redis password is `94ER#p`, I can connect to redis with `redis://:94ER23%p@redis_ip_or_domain` in celery, but I cannot pass the argument to django-health-check. `redis://:94ER#p@redis_ip_or_domain` will be assume as 
`(scheme='redis', netloc='94ER', path='', params='', query='', fragment='p@redis_ip_or_domain')` because `#` is the fragment indicator.

I agree it should be fixed by redis-py, the `decode_components` argument should default to `True`. But redis-py has already fix it in the next 4.0 version. And they will not change the behavior in 3.x because you can pass the argument to implement what you want.

PMExtra:
At least, django-health-check should provides another way to set the redis password in any character.

PMExtra:
@codingjoe What do you think?

codingjoe:
Hm, I don't know. The change in redis makes sense, I believe the change in the library does not. However, development on redis seems to be rather slow. I don't mind adding a backport if you fix whatever broke the tests. Best, Joe

PR #278, opened 2020-11-05
Fix Celery ping health check functionality with multiple workers

### LINKED ISSUE #277: Celery ping check does not work with multiple workers
`app.control.ping()` actually returns `List[Dict[str, Dict[str, str]]]` not a `List` with one `Dict[str, Dict[str, str]]`. https://github.com/KristianOellegaard/django-health-check/blob/master/health_check/contrib/celery_ping/backends.py#L15 works fine with one worker but not with multiple ones. I'll address this issue myself later on today or this week.

maszaa:
Underlying problem was a misinterpretation of app.control.ping return value

Fixes https://github.com/KristianOellegaard/django-health-check/issues/277

### DIFF
--- a/health_check/contrib/celery_ping/backends.py
+++ b/health_check/contrib/celery_ping/backends.py
@@ -35,7 +35,8 @@ class CeleryPingHealthCheck(BaseHealthCheckBackend):
     def _check_ping_result(self, ping_result):
         active_workers = []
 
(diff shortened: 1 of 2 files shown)


codingjoe:
Released in [3.16.1](https://pypi.org/project/django-health-check/3.16.1/) 🎉 

Thanks @maszaa for the hotfix

PR #281, opened 2020-11-30
Add simple endpoint to check if application is initialized

Chris7:
This is a lightweight endpoint that can be used by services to monitor the uptime of an application without going through the full gamut of healthchecks. It's meant for using in something like a k8s cluster where you want a simple endpoint to know an application has been initialized, or to continually probe for uptime.

### DIFF
--- a/health_check/urls.py
+++ b/health_check/urls.py
@@ -1,9 +1,10 @@
 from django.urls import path
 
-from health_check.views import MainView
+from health_check.views import MainView, ping
 
 app_name = 'health_check'
 
 urlpatterns = [
     path('', MainView.as_view(), name='health_check_home'),
+    path('ping', ping, name='health_check_ping'),
 ]
--- a/health_check/views.py
+++ b/health_check/views.py
@@ -110,3 +110,6 @@ class MainView(CheckMixin, TemplateView):
             {str(p.identifier()): str(p.pretty_status()) for p in plugins},
             status=status
         )
+
+def ping(request):
+    return HttpResponse('pong')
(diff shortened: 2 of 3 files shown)


codingjoe:
Hi @Chris7,
Thanks for reaching out, and thank you for your contribution. Code wise, I have no comments, it's pretty straightforward. I am curious about the use case, though. This is very different from what this package usually does. With that in mind, the code you added anyone could easily implement in their applications themselves. With that in mind I am a bit conflicted to merge it.
I'd appreciate multiple opinions. If you find more people who request this feature, I would be happy to merge it.
Best,
Joe

Chris7:
Thanks for your fast response. Here's some more context for our use case:

I agree it is trivial, but for a place with multiple django servers running, having all your healthchecks added in a single package is very useful rather than implementing the same snippet every time.

Here's 2 cases we use this endpoint for:
1) Service uptime -- we have a bot that hits ping every couple of seconds to provide latency and uptime information -- having a long healthcheck or one that hangs on unresponsive services (e.g. s3) can actually end up taking down your application if you overwhelm all the worker threads.
2) Container registration in our docker deployment. A service needs to tell the load balancer that it is ready to accept traffic after a container has been started. This is usually called a "readiness probe" in kubernetes. It means that the application has been initialized and is ready to serve traffic and can be added to the load balancer.

But we can see if anyone else fits that niche as well :).

seanpar203:
I'm going to close this PR as it's not needed and seems specific to a particular infrastructure setup or convenience due to infrastructure.

PR #283, opened 2021-01-06
Added text align right for time col

anushbmx:
The w3-right class does not specific a text alignment, which makes the time values to be misaligned in certain browsers like safari.

so added w3-right-align to for text align to right in last col for time taken.

Note: even without w3-right-align Firefox and chrome aligns the text to right, the issue was noticed in safari. Attached screen shot for reference.

Screenshot 2021-01-06 at 12 12 05 PM

### DIFF
--- a/health_check/templates/health_check/index.html
+++ b/health_check/templates/health_check/index.html
@@ -13,7 +13,7 @@
       <tr class="w3-blue">
         <th colspan="2">Service</th>
         <th>Status</th>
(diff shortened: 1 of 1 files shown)


codingjoe:
Thanks @anushbmx 👍 good catch!

codingjoe:
Released in [3.16.2](https://pypi.org/project/django-health-check/3.16.2/) 🎉

PR #292, opened 2021-04-13
feat: add default_auto_field to db

### LINKED ISSUE #291: Django 3.2 DEFAULT_AUTO_FIELD Warning
[Django 3.2](https://docs.djangoproject.com/en/3.2/ref/settings/#default-auto-field) wants either `DEFAULT_AUTO_FIELD` to be set globally or `default_auto_field` in app config, otherwise the following warning shows up:

```
WARNINGS:
db.TestModel: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the HealthCheckConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.
```

I can open a PR that adds a default to the [app config](https://github.com/KristianOellegaard/django-health-check/blob/master/health_check/db/apps.py#L6-L7) if that works.

tmm:
Closes #291


codingjoe (commented):
Hi @tmm. Why would you want to change it to a big integer? I think just adding an explicit value should be enough. This table is usually empty, if someone really runs out of integers you can always reset the sequence.

tmm:
@codingjoe for some reason I thought that Django defaulted to `BigAutoField`. I'll update to `AutoField` instead.

codingjoe:
Thanks @tmm. Released in [3.16.4](https://github.com/KristianOellegaard/django-health-check/releases/tag/3.16.4) 🎉

Issue #290, opened 2021-04-09
Access to Django request from within the plugin


db4y:
Hi

I'm writing a custom check and I would need some informations contained in Django's request object in my check. I could not find a way to access the request nor to pass it to the plugin.

Is it possible to access the request from the plugin?

Thank you.

skarzi:
Hi :wave:

I'm afraid it's not possible because [`CheckMixin`](https://github.com/KristianOellegaard/django-health-check/blob/master/health_check/mixins.py#L9) that is subclassed in [`MainView`](https://github.com/KristianOellegaard/django-health-check/blob/master/health_check/views.py#L80) doesn't pass the `request` object down to plugins `run_check` calls.
So the only way is to rewrite this package flow or (personally I dislike this solution) use some middleware to set currently handled `request` to some global variable

db4y:
Thank you for the quick answer @skarzi 
Do you think it would be possible to add this feature to the package? I would be willing to help.

codingjoe:
Yes, it would be 😉  If you want to contribute this feature I'd be happy to review it. Best, Joe

codingjoe:
Hi there 👋, I will close this due to inactivity. Please don't hesitate to reopen this issue should there be sustained interest in the feature. Cheers, Joe

PR #294, opened 2021-04-21
Inject context dict to plugin if it has a set_context method

db4y:
Hi @codingjoe 

About #290 I came up with this.

If you need to inject something to your plugin you would

1. subclass the view and create a `get_plugin_context(plugin_name)` in it. In that method you can inject what you need. Since the plugin name is passed as an argument, you can customize what you inject for each plugin.
2. in your plugin you can now access `self.context` and override the `set_context()` method if needed.

What do you think?

### DIFF
--- a/health_check/backends.py
+++ b/health_check/backends.py
@@ -19,6 +19,7 @@ class BaseHealthCheckBackend:
 
     def __init__(self):
         self.errors = []
+        self.context = {}
 
     def check_status(self):
         raise NotImplementedError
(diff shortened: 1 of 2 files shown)


codingjoe (commented):
Thanks @db4y,

this looks promising. I just had a first glance to be able to leave you simple architecture feedback. I understand the context approach, but that's probably more than is needed here. Maybe we keep it simple and just add the `request` as an argument to `run_check`.
I don't mind a breaking release. It's easily fixed, should people use their custom checks.

Best,
Joe

db4y:
I get your point @codingjoe I thought the context solution would cover hypothetical future needs and allow to pass any object that the plugin may need from the view.

seanpar203:
I'm closing this due not being needed.

Maybe a good abstraction and way to make things more robust or powerful, but not necessarily needed.

PR #301, opened 2021-06-04
Added simple health check using a list of callable endpoints via HTTP GET

josemlp91:
I think this feature could be interesting when your Django Application depends on several microservices or third party services and to prevent network issues.

Let me know your opinion about this feaure, and if a candidate to merge in master.

### DIFF
--- a/health_check/contrib/http_ping/backends.py
+++ b/health_check/contrib/http_ping/backends.py
@@ -24,15 +24,22 @@ class HTTPPingHealthCheck(BaseHealthCheckBackend):
         logger.debug("Got %s as the http_endpoints. ", http_endpoints)
 
         for http_endpoint in http_endpoints:
+            
             try:
-                requests.get(http_endpoint, timeout=timeout)
+                response = requests.get(http_endpoint, timeout=timeout)
+
             except Timeout as e:
                 self.add_error(ServiceUnavailable(f"Timeout conecting to {http_endpoint}"), e)
                 return
+
             except Exception as e:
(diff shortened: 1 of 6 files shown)


health_check/contrib/http_ping/backends.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def check_status(self):
+        """Check http conections for all urls in settings.HTTP_ENDPOINTS_HEALTH_CHECK"""
+
+        http_endpoints = getattr(settings, "HTTP_ENDPOINTS_HEALTH_CHECK", [])
+        timeout = getattr(settings, "HTTP_TIMEOUT_HEALTH_CHECK", 1)
+
+        logger.debug("Got %s as the http_endpoints. ", http_endpoints)
+
+        for http_endpoint in http_endpoints:
```
anudeepsamaiya, line 26: what are your opinions on using multiple threads for each endpoint, somthing like:

```
    def check_status(self):
        errors = []

        def _check_status(http_endpoint):
            return requests.get(http_endpoint, timeout=timeout)

        with ThreadPoolExecutor(max_workers=len(self.http_endpoints) or 1) as executor:
            for http_endpoint in executor.map(_check_status, self.http_endpoints):
                if self.critical_service:
                    if not HEALTH_CHECK["WARNINGS_AS_ERRORS"]:
                        self.errors.extend(
                            e
                            for e in plugin.errors
                            if not isinstance(e, ServiceWarning)
                        )
                    else:
                        self.errors.extend(plugin.errors)

        return errors

```

PR #305, opened 2021-08-06
Improve test reliability by cleaning the state pollution of test_warning in TestMainView

sturmianseq:
This PR aims to improve test reliability by cleaning the state pollution in `TestMainView.test_warning`. `HEALTH_CHECK['WARNINGS_AS_ERRORS']` is set to `TRUE` by default, but `test_warning` changes it to be `FALSE`. The fix is to restore the default value after running the test.

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -99,7 +99,7 @@ class TestMainView:
         assert response['content-type'] == 'text/html; charset=utf-8'
         assert b'Super Fail!' in response.content
 
-    def test_warning(self, client):
+    def test_warning(self, client, monkeypatch):
         class MyBackend(BaseHealthCheckBackend):
             def check_status(self):
                 raise ServiceWarning('so so')
@@ -110,7 +110,7 @@ class TestMainView:
         assert response.status_code == 500, response.content.decode('utf-8')
         assert b'so so' in response.content, response.content
 
-        HEALTH_CHECK['WARNINGS_AS_ERRORS'] = False
+        monkeypatch.setitem(HEALTH_CHECK, 'WARNINGS_AS_ERRORS', False)
 
         response = client.get(self.url)
         assert response.status_code == 200, response.content.decode('utf-8')


tests/test_views.py
```diff
@@ -116,6 +116,7 @@ def check_status(self):
         assert response.status_code == 200, response.content.decode('utf-8')
         assert response['content-type'] == 'text/html; charset=utf-8'
         assert b'so so' in response.content, response.content
+        HEALTH_CHECK['WARNINGS_AS_ERRORS'] = True
```
skarzi: this will evaluate only when all `assert` statements will pass.
`pytest` has neat fixture [`monkeypatch`](https://docs.pytest.org/en/6.2.x/monkeypatch.html#monkeypatching-dictionaries) that can be used to override `HEALTH_CHECK` dictionary just for one test. Just use `monkeypatch.setitem` to update `WARNINGS_AS_ERRORS` key:


```python
monkeypatch.setitem(HEALTH_CHECK, 'WARNINGS_AS_ERRORS', False)
````

Comment (skarzi) [reply tests/test_views.py]: This project also uses `pytest-django` that has great fixture - `settings` that can be also used to modify `django-health-check` settings:

```python
settings.HEALTH_CHECK['WARNINGS_AS_ERRORS'] = True
```

And personally, I think it's the best way to write this test because we are already using `pytest-django`'s `client` fixture.

sturmianseq: Thanks a lot for your comments! `monkeypatch` is a great solution! I strongly agree that your fix is much better than my original one.

By the way, I tried to use `settings` but it failed in the following way.
```
tests/test_views.py:113: in test_warning
    settings.HEALTH_CHECK['WARNINGS_AS_ERRORS'] = True
/home/sturmian/.local/lib/python3.8/site-packages/pytest_django/fixtures.py:347: in __getattr__
    return getattr(settings, item)
/home/sturmian/.local/lib/python3.8/site-packages/django/conf/__init__.py:83: in __getattr__
    val = getattr(self._wrapped, name)
E   AttributeError: 'Settings' object has no attribute 'HEALTH_CHECK'
```

Please let me know if you would like to improve my current fix, thanks!

skarzi (approved):

tests/test_views.py
skarzi: It seems like `django-health-check` is not initialized in `tests/test_app/settings.py`, so let's stick to `monkeypatch` :+1:

seanpar203:
Closing due to lacking details or reliable test steps for reproducibility.

PR #312, opened 2021-11-16
Prometheus Metrics View implementation

sh0ked:
This PR adds support for getting health check results in Prometheus format. See https://prometheus.io/.
You can also add new metrics to basic health checks or to your own.
For the implementation of prometheus metrics, we use the official package https://github.com/prometheus/client_python.


nniehoff:
@KristianOellegaard any chance we could get this implemented?

MartinSchmidt123:
I would also support adding prometheus metric exports.

KristianOellegaard:
I think its a great idea, but it shouldn't be a requirement to install django-health-check, but rather be an optional requirement for the ones who want to use it. As I read the current code, everyone needs to install it.

denisSurkov:
Hi!

I thinks this is really greate feature for django-healthchecks. 

Any chance someone can continue this PR? 

@KristianOellegaard, can you help to make this optional (share idea how to do it).
@frankwiles, can you check it or share idea how to make it optional? (Sorry, I mentioned you because you are the one who made the last commit)

frankwiles:
Oh wow I hadn't seen this PR yet @denisSurkov I think it's a great idea, but agree it needs to be optional. The metrics client needs Twisted which is a fairly heavy dependency to require.  

I think we'd want to adjust this to conditionally import the client, warn if `USE_PROMETHEUS` is set but client isn't installed, etc. 

But overall I think a decent idea.  We typically have a separate Prometheus endpoint, but I don't see any real reason it couldn't be the same endpoint. 

We just took over this module.  I'm currently reworking the docs and getting familiar with the code base, GH Actions, etc. but I think this might be the next most important thing on the list.

denisSurkov:
@sh0ked , can you continue your pull request?

@frankwiles , sounds cool. If the PR author would not respond, I would like to help to implement this feature.

sh0ked:
Hi guys! Unfortunately, I don't have time to continue this PR. But I will be glad if someone can improve it.

codingjoe:
Howdy, y'all 👋,

Since the PR is abandoned, I will take the liberty to close the PR.

In addition, please note #385

Best!
Joe

PR #318, opened 2022-01-07
Remove the need for external CSS

RealOrangeOne:
This reduces unnecessary external requests when viewing the endpoint in a browser.

This intentionally uses inline CSS to avoid an extra query to Django. Could be moved to an external template and included if needed

### DIFF
--- a/health_check/templates/health_check/index.html
+++ b/health_check/templates/health_check/index.html
@@ -36,7 +36,7 @@
       text-align: right;
     }
 
-    table .header {
+    table thead {
       background-color: lightskyblue;
     }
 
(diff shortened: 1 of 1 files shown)


codingjoe:
@RealOrangeOne thanks for the contribution. I understand why you'd want to change that. Would you be so kind as to rebase your PR and to include a screenshot in your PR description? Ping me, once you're done, and I'll merge this. Cheers, Joe

health_check/templates/health_check/index.html
```diff
@@ -2,31 +2,73 @@
 <head>
   <title>{% block title %}System status{% endblock title %}</title>
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <link rel="stylesheet" href="https://www.w3schools.com/w3css/4/w3.css">
-  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
+  <style>
+    html, body {
+      margin: 0 auto;
+      font-family: Verdana, sans-serif;
```
codingjoe: We use two different fonts, is that on purpose? Maybe we just default to the system font. Perhaps even a monospaced font?

health_check/templates/health_check/index.html
```diff
@@ -2,31 +2,73 @@
...
+
+    td, th {
+      padding: 8px;
+    }
+
+    .align-right {
+      text-align: right;
+    }
+
+    table .header {
```
codingjoe: Maybe we introduce proper `thead` tags?

health_check/templates/health_check/index.html
RealOrangeOne: It wasn't, no. I think I just borrowed it from the old styles. I've updated it to use just 1 font, the system `sans-serif` font.

RealOrangeOne:
@codingjoe thanks - over to you!

codingjoe (approved):

codingjoe:
Thanks for the great contribution @RealOrangeOne 
Your change is released in 3.17 🎉

PR #327, opened 2022-07-29
Add proper line breaks to the pretty_status HTML

kbyun03:
since `pretty_status` returns linebreak-ed string, I think it makes sense to add `linebreak` in the default template.


codingjoe (approved):

codingjoe:
Thanks, @kbyun03, excellent suggestion. Cheers, Joe

PR #335, opened 2022-12-05
Use default number of `max_workers`

RealOrangeOne:
This allows Python to decide what the right number is, based on the number of CPU cores available.


seanpar203:
Going to talk to Frank about this and circle back.

seanpar203:
Looks like this has already been included as part of the code.

Going to close this since it's already done.

PR #336, opened 2022-12-07
CacheBackend: Use a configurable setting to specify cache key

Flauschbaellchen:
Using a static class variable allows applications using django-health-check as a 3rd-party package to read the key's identifier used to read/write the cache entry.

This is in particular useful, if the application needs to know which keys are written from other packages and are not part of the application logic itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,13 @@ on django.conf.settings with the required format to connect to your redis server
     REDIS_URL = redis://localhost:6370
 ```
 
+The cache healthcheck tries to write and read a specific key within the cache backend.
+It can be customized by setting `HEALTHCHECK_CACHE_KEY` to another value:
+
+```python
+    HEALTHCHECK_CACHE_KEY = custom_healthcheck_key
+```
+
 ## Setting up monitoring
 
 You can use tools like Pingdom, StatusCake or other uptime robots to monitor service status.
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -71,7 +71,7 @@ exceeds 90% or available memory drops below 100 MB.
    memory falls below the specified value, a warning will be reported.
 
 Celery Health Check
-----------------------
+-------------------
 Using `django.settings` you may exert more fine-grained control over the behavior of the celery health check
 
 .. list-table:: Additional Settings
@@ -82,6 +82,10 @@ Using `django.settings` you may exert more fine-grained control over the behavio
      - Type
      - Default
      - Description
+   * - `HEALTHCHECK_CACHE_KEY`
+     - String
+     - djangohealtcheck_test
+     - Specifies the name of the key to write to and read from to validate that the cache is working.
    * - `HEALTHCHECK_CELERY_QUEUE_TIMEOUT`
      - Number
      - 3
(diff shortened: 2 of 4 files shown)


health_check/cache/backends.py
```diff
@@ -5,6 +5,10 @@
 
 
 class CacheBackend(BaseHealthCheckBackend):
+
+    # Cache key used to write and read against
+    HEALTH_CHECK_KEY = "djangohealtcheck_test"
```
skarzi: Why not make it configurable via some Django setting e.g. `HEALTH_CHECK = {'CACHE_CHECK_KEY': 'my-cache-check-key'}`?

Flauschbaellchen: @skarzi well, your right. My requirement was to _read_ the key's name, but changing it to another custom one would might solve other problems, too.
Also, if setting the key's name within the Django settings would still allow the developer to access it, so my requirement would still be solved :+1: 

I've updated my PR.

Flauschbaellchen: I've choosen to use `HEALTHCHECK_CACHE_KEY` instead of extending `HEALTH_CHECK`.
It seems that `HEALTH_CHECK` is only used for `psutils` and other "single" settings exists (e.g. `HEALTHCHECK_CELERY_PING_TIMEOUT`.
I hope that was the right decision.

README.md
```diff
@@ -111,6 +111,13 @@ on django.conf.settings with the required format to connect to your redis server
     REDIS_URL = redis://localhost:6370
 ```
 
+The cache healthcheck tries to write and read a specific key within the cache backend.
+It can be customized by setting `HEALTHCHECK_CACHE_KEY` to another value:
+
+```python
+    HEALTHCHECK_CACHE_KEY = custom_healthcheck_key
```
skarzi, line 118:
```suggestion
    HEALTHCHECK_CACHE_KEY = 'custom_healthcheck_key'
```

docs/settings.rst
```diff
@@ -82,6 +82,10 @@ Using `django.settings` you may exert more fine-grained control over the behavio
      - Type
      - Default
      - Description
+   * - `HEALTHCHECK_CACHE_KEY`
+     - String
+     - djangohealtcheck_test
```
skarzi, line 87:
```suggestion
     - 'djangohealtcheck_test'
```

skarzi (commented):
nice work 👍

README.md
Flauschbaellchen: @skarzi For both suggestions I was about to add those quotes around them, but as they were also left out from the other settings, I decided to leave them as they are.

Maybe this would be another refactoring PR after merging this one? In this way we could fix the other settings, too.

skarzi: ok, so then it makes sense to add these quotes in a separate PR 👍

Flauschbaellchen: @skarzi What are the next steps here? Do you resolve the discussions in case you're fine with my decisions or am I allowed to do that? In my opinion all threads can be resolved/closed.

Flauschbaellchen:
@skarzi Any news on this? I've resolved the threads as no more work needs to be done.

skarzi:
@Flauschbaellchen I am not a maintainer of this package, just a contributor. I think this PR could be merged (and thank you - I am running your changes from this PR on prod in some of my projects for ~1 month 👍).

@frankwiles can you take a look at this PR?

frankwiles (approved):
Looks great.  Thanks!

PR #351, opened 2023-01-17
Respect celery configuration namespace

ercpe:
When configuring Celery with RabbitMQ in Django, one often uses `app.config_from_object('django.conf:settings', namespace='CELERY')` as per the Celery Quick Start (https://docs.celeryq.dev/en/stable/django/first-steps-with-django.html). This requires that all configuration keys in the Django settings have to be prefixed with `CELERY_`.

The previous implementation only used `settings.BROKER_URL` (which may be `None`). When passing `None` to Celery (or more specifically: the `Connection` object), it would default to `localhost`. Expand the `RabbitMQHealthCheck` to accept a `namespace` attribute which can be set to `CELERY` for example.


frankwiles:
Looks good to me.  Thanks!

PR #357, opened 2023-02-27
Streamline quotes in documentation

Flauschbaellchen:
This is a follow-up from #336.
This PR streamlines and corrects quotes around strings and inline code.


frankwiles:
Great thanks!

PR #359, opened 2023-03-17
Improve configuration of default storage

### LINKED ISSUE #366: Provide a fix to avoid " The DEFAULT_FILE_STORAGE setting is deprecated"
Hello, thanks for a great app!

I used django-health-check for a while with no problems.
But recently I faced with the issue when I tried to execute a command `poetry run pytest tests`: 
django.utils.deprecation.RemovedInDjango51Warning: The DEFAULT_FILE_STORAGE setting is deprecated. Use STORAGES instead.
note: Django 4.2 introduces the new [STORAGES setting](https://docs.djangoproject.com/en/4.2/releases/4.2/#custom-file-storages).
Could you fix this issue

ron8mcr:
Previously it was set using `str` with dotted path to storage, so `StorageHealthCheck.get_storage` initialize it on every call. This may lead to redundant calls to external APIs (see https://github.com/boto/boto3/issues/3614)

Meanwhile `django.core.files.storage.default_storage` is lazy object that can be safely used as class attribute, and single `storage` object may be reused

Also fixes #366


vanya909:
@frankwiles Hello, can you please check this PR? Just curious about will I be able to use this feature soon

ron8mcr:
Hi @frankwiles, could you please take a loot at this one or at #408 or #383 ?
Without this fix, `django-health-check` cannot be used with newer versions of Django (see https://docs.djangoproject.com/en/dev/releases/5.1/#features-removed-in-5-1)

PR #362, opened 2023-03-30
Add configuration for running health checks without threads

mnitchie:
Addresses https://github.com/revsys/django-health-check/issues/360

A configuration option to allow for running health checks without spinning up threads would be useful in a few different scenarios.

This adds a HEALTH_CHECK["DISABLE_THREADING"] config to support this. The config defaults to `False` so existing consumers of the lib will still work the same way.

h/t @actionscripted


dimaqq:
@ilyamonakov review pls

rapkyt:
Can this be merged? I have randoms time outs only for the health check endpoint that I think it will be solved by not using threading, thanks!

mnitchie:
@frankwiles and/or @jefftriplett - thanks for your work on this repo! Curious what you think of this and if it could be merged.

frankwiles:
Yeah I think we can merge this.  I didn't even remember it was using threads, but can definitely see that getting in the way. 

My plan/hope for v2 (when I can find the time for it) is for there to be a separate view that will run the checks async for a nice UI/UX display, but keep them sync obviously for automated use cases (i.e. load balancers knowing which are healthy, etc)

frankwiles (approved):

frankwiles:
I'll get a release put together in a couple of days.

ron8mcr:
> I'll get a release put together in a couple of days.

@frankwiles Hope that https://github.com/revsys/django-health-check/pull/397 will be included to this release too

PR #364, opened 2023-04-12
Update cache/backends.py

sergio-bershadsky:
Check status produces exception when service (ex. Redis) is not available. Thus it makes django output 500 error page instead of pretty printing the cause of error.


seanpar203:
Closing due to lack of details, clarity, and tests.

PR #365, opened 2023-05-02
refactor(django50): replace DEFAULT_FILE_STORAGE wiht STORAGES setting

lennarddevries:
RemovedInDjango51Warning: The DEFAULT_FILE_STORAGE setting is deprecated. Use STORAGES instead.


Kub-AT:
How about backward compatibilities?

PR #370, opened 2023-07-06
Allow selecting which celery worker(s) to ping

sevdog:
This PR aims to add the ability to select which celery worker needs to be ping-ed.

The main use case is to use the management command to check a specific worker, thus the check should be ok only if _that_ worker is up&running. Since celery [`app.control.ping`](https://docs.celeryq.dev/en/stable/reference/celery.app.control.html#celery.app.control.Control.ping) allows a list as parameter the settings allows a list too. By default it was kept `None` since it means "ping every worker", which is the current behaviour.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -117,3 +117,26 @@ Using `django.settings` you may exert more fine-grained control over the behavio
      - Number
      - `None`
      - Specifies the healthcheck task priority.
+
+
+Celery-Ping Health Check
+------------------------
+
+Using `django.settings` you may exert more fine-grained control over the behavior of the celery-ping health check
+
+.. list-table:: Additional Settings
+   :widths: 25 10 10 55
+   :header-rows: 1
+
+   * - Name
+     - Type
+     - Default
+     - Description
+   * - `HEALTHCHECK_CELERY_PING_TIMEOUT`
+     - Number
+     - `1`
+     - Specifies the maximum total time (in seconds) for which "pong" responses are awaited.
+   * - `HEALTHCHECK_CELERY_PING_DESTINATION`
+     - List of Strings
+     - `None`
+     - Specifies the list of workers which will receive the "ping" request.
--- a/health_check/contrib/celery_ping/backends.py
+++ b/health_check/contrib/celery_ping/backends.py
@@ -10,9 +10,10 @@ class CeleryPingHealthCheck(BaseHealthCheckBackend):
 
     def check_status(self):
         timeout = getattr(settings, "HEALTHCHECK_CELERY_PING_TIMEOUT", 1)
+        destination = getattr(settings, "HEALTHCHECK_CELERY_PING_DESTINATION", None)
 
         try:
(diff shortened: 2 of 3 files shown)


codingjoe:
Hi @sevdog 👋,

Thank you for your contribution and your patience. I am partially hesitant to add fairly use-case-specific settings. Inheritance is an excellent approach to customizing a check to your needs. If the API is not extensible enough or we require more documentation, I am happy to accept those patches any day.

As for this one, I am closing the PR for now. Please don't hesitate to reopen it and to make your case for mainline support.

Happy holidays!
Joe

sevdog:
Hi @codingjoe, I am in favor of using inheritance when possible, yet in this case it would require a full rewrite of the `CeleryPingHealthCheck` class to achieve the same goal. This because what this PR aims is to allow to ping only the desired celery worker for two reason:

- Do not bother any other worker in which we are not interested
- Do not have to wait for eventual errors (KO or timeout) from other workers

With this PS the check is still configurable to hold both the full-broadcast and the targeted use cases. However if you prefer having two different checkers I am fine to open an other PR in which I can refine the current API of `CeleryPingHealthCheck` to build a `CeleryHostPingHealthCheck` which inherits from it; thus having the project with two _HC_ which can manage celery ping.

codingjoe:
Hi,

Thanks for getting back to me so quickly. Looking at your code, you'd currently only need to override `_check_ping_result` to selectively ping workers.

Since Celery is only part of the contrib section of this package, since it isn't Django. I am hesitant to accept new features, especially with tasks in Django 6.

I would encourage you to open an issue to see if there is wider community interest.

Cheers,
Joe

sevdog:
Hi @codingjoe 
> Thanks for getting back to me so quickly. Looking at your code, you'd currently only need to override `_check_ping_result` to selectively ping workers.

This is not true, because just overring that method will not satisfy my two requirements:

> - Do not bother any other worker in which we are not interested ❌ 
> - Do not have to wait for eventual errors (KO or timeout) from other workers ❌

Because the ping will reach all workers without providing the `destination` argument in `.check_status` method.
As stated in [app.control.ping](https://docs.celeryq.dev/en/stable/reference/celery.app.control.html#celery.app.control.Control.ping):
> Ping all (or specific) workers.
```python
app.control.ping()
# [{'celery@node1': {'ok': 'pong'}}, {'celery@node2': {'ok': 'pong'}}]
app.control.ping(destination=['celery@node2'])
# [{'celery@node2': {'ok': 'pong'}}]
```

So the change in the invocation of `app.control.ping` is mandatory for this feature to work as expected (without having to rewrite the whole class).

I understand that changes in django core are going to alter also this kind of check, yet this PR was made years before that change took place and does not use any hidden or undocumented APIs of celery (which is the real focus for this checker).

codingjoe:
@sevdog, I hear you. Please open a new ticket. I am currently reworking the architecture in a larger breaking release. I will facilitate the ability to add settings per check. This will simplify adding new settings and functionality like you are proposing.

PR #378, opened 2023-09-08
Update README.md

chadgates:
Replace url with path. Assume that was the source for https://github.com/revsys/django-health-check/issues/373


jefftriplett:
@chadgates thank you for the contribution!

PR #390, opened 2023-11-25
Add ability to define and run a subset health checks

panteparak:
This PR Introduces an ability to define and run a `SUBSET` of health checks.

It will be useful during certain period where it is only required to run specific health checks during a scenario.

It supports both via url `/ht/<subset-name>/` and via Django command `python manage.py health_check --subset <subset-name>`


panteparak:
Hi @codingjoe 

Just a polite ping!. 

Could I have some feedback on this PR. 

Really love to see this feature in the official release of django-health-check.

frankwiles (approved):
This looks great @panteparak I think it's definitely worth including.  I'm hoping to do another release in the next few days (was actually supposed to do it last weekend but something came up) so I can include this in there.

panteparak:
@frankwiles Just a friendly reminder: Don't forget to create a new release :)

saz:
@frankwiles FYI: this is a breaking change, if you've got custom checks, as the signature of `BaseHealthCheckBackend.check_status` has changed.
```
TypeError: CeleryBeatCheckBackend.check_status() got an unexpected keyword argument 'subset'
```
I'll send a PR to fix the docs. Maybe it's possible to add some hint to the current release notes?

SpecLad:
It's not clear to me what the point of the `subset` parameter even is. None of the builtin health check backends actually use it. Can't it be removed to restore compatibility?

ramwin:
> @frankwiles FYI: this is a breaking change, if you've got custom checks, as the signature of `BaseHealthCheckBackend.check_status` has changed.
> 
> ```
> TypeError: CeleryBeatCheckBackend.check_status() got an unexpected keyword argument 'subset'
> ```
> 
> I'll send a PR to fix the docs. Maybe it's possible to add some hint to the current release notes?

I agree with you, this merge request should be merged into 4.0.0 only.

saz:
@SpecLad There's an example usage in the README (not within the docs).

SpecLad:
> @SpecLad There's an example usage in the README (not within the docs).

Where? The README hasn't even been updated to _include_ the `subset` parameter.

saz:
image
Can't link to the proper section, but as said, it's only in the README, not the documentation.

SpecLad:
This doesn't describe the `subset` parameter.

saz:
Ah, sorry, misunderstanding on my side. It looks like as if the `subset` parameter on `check_status` isn't used anywhere.

@panteparak Can you shed some light on this?

llam15:
I would like to +1 that this is a breaking change and should not have been released as a minor version update. This caused all of our custom health checks to fail with a `TypeError`
> ...check_status() got an unexpected keyword argument 'subset'

The minor version upgrade from `3.17` to `3.18` was pulled automatically because we had specified our dependency `django-health-check = "^3.16"`, and unexpectedly caused the health checks to stop working.

---
For others: we were able to fix our custom health checks by modifying the `check_status` definition as so:
 ```
def check_status(self, *args, **kwargs):
```

SpecLad:
I made a PR to unbreak this: #414.

PR #414, opened 2024-01-30
Fix compatibility with pre-3.18 custom check backends

### LINKED ISSUE #413: Bug: check_status() fails after upgrading to 3.18
Hi all,

I just want to make you aware of an issue I've encountered after upgrading from 3.17 to 3.18:

`CustomCheck.check_status() got an unexpected keyword argument 'subset'`

This error has been raised for some of our custom health checks. I suppose it's related to this PR? https://github.com/revsys/django-health-check/pull/390

From my perspective, the new feature doesn't work with the example provided in the Readme.

SpecLad:
The `subset` parameter added to `BaseHealthCheckBackend`'s `run_check` and `check_status` methods in 4da1fd9d broke compatibility with existing custom check backends, since they don't accept this parameter.

Moreover, this parameter appears to be entirely useless, since `check_status` is only called by `run_check`, and `run_check` is only called by `CheckMixin.run_check`, which never passed a value for `subset`.

Get rid of it and restore compatibility.

Fixes #413.

### DIFF
--- a/health_check/backends.py
+++ b/health_check/backends.py
@@ -20,14 +20,14 @@ class BaseHealthCheckBackend:
     def __init__(self):
         self.errors = []
 
-    def check_status(self, subset=None):
+    def check_status(self):
         raise NotImplementedError
 
-    def run_check(self, subset=None):
+    def run_check(self):
         start = timer()
         self.errors = []
         try:
-            self.check_status(subset=subset)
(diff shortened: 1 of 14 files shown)


frankwiles:
Thanks for this! Great catch, not sure how we missed this!

PR #421, opened 2024-03-19
Allow Configuration of Test Folder Path

striveforbest:
Description:

This pull request addresses an issue where the library attempts to create a folder named health_check_storage_test within the current directory. However, in many setups, users may lack the necessary permissions to do so. To resolve this, the pull request introduces a feature that enables users to easily configure the folder/path through a setting.

Changes:

    Added support for configuring the folder/path using a setting.

    Users can now define the desired folder/path in their Django settings by adding the following line:

```python
HEALTH_CHECK_TEST_FOLDER = '/tmp'
```

This enhancement offers greater flexibility and ease of use for users, allowing them to specify a folder/path that suits their setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,14 @@ The backend will return a JSON response:
     }
 ```
 
+## Settings overrides
+
+You can provide `settings.HEALTH_CHECK_TEST_FOLDER` to alter the default storage behavior. It is helpful if your user doesn't have permissions to create a folder within the current dir.
+
+```python
+HEALTH_CHECK_TEST_FOLDER = '/tmp'
+```
+
 ## Writing a custom health check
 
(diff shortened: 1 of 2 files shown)


striveforbest:
Oh, this won't work. I'll work on a different solution.

seanpar203:
Talking with Frank about this, will circle back!

seanpar203:
Closing this as the `get_file_name` is actually mocked within the tests:


```py

def get_file_name(*args, **kwargs):
    return "mockfile.txt"

@mock.patch(
    "health_check.storage.backends.StorageHealthCheck.get_file_name", get_file_name
)
```

So I don't think this is needed anymore.

PR #431, opened 2024-07-05
Add support for database specific health checks

HanaPoulpe:
Enables parameterised `DatabaseBackend` with specific database name like `DatabaseBackend[secondary]`.

Larger Django application can use specific databases to handle specific part of the application, some database might be required at application start-up while other are only used by non-vital part of the application.

This change allow a database per database health check configuration.

Part 1 of 3 for [Support multi Database Health Checks](https://github.com/revsys/django-health-check/issues/430)

### DIFF
--- a/health_check/db/apps.py
+++ b/health_check/db/apps.py
@@ -1,4 +1,5 @@
 from django.apps import AppConfig
+from django.conf import settings
 
 from health_check.plugins import plugin_dir
 
@@ -11,3 +12,6 @@ class HealthCheckConfig(AppConfig):
         from .backends import DatabaseBackend
 
         plugin_dir.register(DatabaseBackend)
+
+        for database_name in settings.DATABASES.keys():
+            plugin_dir.register(DatabaseBackend, database_name=database_name)
--- a/README.md
+++ b/README.md
@@ -103,6 +103,20 @@ To use Health Check Subsets, Specify a subset name and associate it with the rel
     }
 ```
 
+To add checks on a specific database, it's possible to parameterize `DatabaseBackend` to use a specific database:
+```python
+    HEALTH_CHECK = {
+        # .....
+        "SUBSETS": {
+            "database-probe": [
+                "DatabaseBackend[default]",  # This is equivalent to "DatabaseBackend"
+                "DatabaseBackend[secondary]",
+            ],
+        },
+        # .....
+    }
+```
+
 To only execute specific subset of health check
 ```shell
 curl -X GET -H "Accept: application/json" http://www.example.com/ht/startup-probe/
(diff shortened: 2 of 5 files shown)


seanpar203:
Talking with Frank about this, will circle back

RemiDesgrange:
Would really appreciate this change (I was going to write a specific backend for it 😄 ). @seanpar203 do you think this PR is ready?

seanpar203:
@RemiDesgrange I'm sorry, i've been torn in several directions.

If this PR wasn't closed yet it's because Frank wanted to add support for it, I discussed with him the Open PR's and we closed the ones that we thought shouldn't have been done.

I think I have to review the PR a little further to see if it works as expected and maybe check out into the branch.

Give me a little time but I will get this in.

RemiDesgrange:
No pressure ✌️

seanpar203 (approved):

health_check/db/apps.py
```diff
@@ -11,3 +12,6 @@ def ready(self):
         from .backends import DatabaseBackend
 
         plugin_dir.register(DatabaseBackend)
+
+        for database_name in settings.DATABASES.keys():
+            plugin_dir.register(DatabaseBackend, database_name=database_name)
```
ddelange, line 17: Hi 👋 

This is a breaking change which caused our health checks to start failing when we upgraded django-health-check. Can you make this change opt-in?

I tried to understand the docs this PR adds to README, but when I search the code base for `database-probe`, there are zero hits (also this diff implies there is no way to opt out). Either way the docs are not clear to me on this breaking change and how to opt out. I would expect opt-in behaviour with docs saying something like "By default, only the `default` database is health checked. To check additional databases, add the following to your settings: ..."

I imagine this change will not only bereak on dummy backend, but also on the read-only replica, since the db health check attempts a write.

Many thanks!

Our settings:
```py
DATABASES = {
    "default": {
        "ENGINE": "django.db.backends.postgresql",
        ...
    },
    # read-only replica
    "replica": {
        "ENGINE": "django.db.backends.postgresql",
        ...
    },
    # virtual database for load balancing - router handles actual routing
    "balanced": {"ENGINE": "django.db.backends.dummy"},
}
DATABASE_ROUTERS = ["app.routers.LoadBalancedRouter"]
```
The stack trace that was introduced by this change:
image

ddelange: opened https://github.com/revsys/django-health-check/issues/500

codingjoe:
As mentioned in #471 and #500 this is causing an issue. Furthermore, the documentation points to behavior that isn't implemented.

Most importantly, Django uses DB routers to read and write operations (and more). We wouldn't force write operations on ALL databases. We can consider adding support to all DATABASES to `DatabaseHeartBeatCheck` instead.

@seanpar203, in conclusion, the change will be reverted and removed as a bug fix release.

Sorry for the inconvenience, everyone. A feature should also not have been released as a micro release. Our bad.

PR #435, opened 2024-07-23
Raise HealthCheckException directly from BaseStorageBackend.check_status

scottgigante:
Currently, all HealthCheckExceptions are raised as "unknown exception" from the storage healthcheck, even if they are, indeed, known exceptions. This PR fixes that.


seanpar203:
Talking with Frank about this, will circle back to you soon.

codingjoe:
Moments GIF

codingjoe:
While merging in the upstream changes, I noticed that this has in fact already been patched in cdb1635c2308371876f5556f03fa169f08b559d7

Thanks again, @scottgigante, for the contribution, and thank you for your patience.

PR #438, opened 2024-10-11
Only close DB connections when using threading

RealOrangeOne:
Previously, connections would be closed when threading was disabled. If opening a connection isn't a cheap operation, this adds latency to both the healthcheck request and future requests processed by the worker.


seanpar203:
Talking with Frank about this, will circle back.

seanpar203:
Support for this was just merged into master.

RealOrangeOne:
@seanpar203 I'm not sure that's true - looking at the code in `master` it still tries to close connections even when unthreaded?

seanpar203:
@RealOrangeOne Sorry, I meant to put that in your other ticket about multiple DB support.

codingjoe (approved):
Hi @RealOrangeOne,

how' you doing? Yes, that change makes a lot of sense. We shouldn't force close connections forcefully. Django will handle them properly if we don't create new threads.

I added some tests to prevent regression.

Cheers!
Joe

PR #439, opened 2024-10-16
feat: add more light weight database check

panteparak:
I noticed that current `health_check.db` create a record and immediately remove such records which is causing some performance degradation on our database. So I've created a custom check that simply performs `SELECT 1;`


panteparak:
Hi @codingjoe and @frankwiles, 
i don't mean to be rude by pinging both of you directly. but, How do get someone to review my PR.

bartvanandel:
I'm not a known party in this repo, but looks good to me. If it were up to me, this could simply be the replacement of the current db check. I don't know why inserting and deleting something is necessary, also, currently it requires a migration to get the db check up. Your replacement doesn't, which IMHO is better.

frankwiles (approved):
Looks good to me @panteparak sorry it took us so long to review and merge.  We'll cut a new release soon FYI.

PR #451, opened 2025-03-20
Remove test model default permissions

### LINKED ISSUE #443: Remove TestModel default permissions
Since default permissions for `TestModel` is not used in tests or views it would be nice to explicitly remove them so they don't show up in admin panel.
Currently installing `health_check.db` in `INSTALLED_APPS` adds 4 default permissions and shows them in admin panel with somehow generic yet ambiguous names:

`db | test model | Can view test model` 
`db | test model | Can add test model` 
`db | test model | Can change test model` 
`db | test model | Can delete test model`

aabharya:
Fixes #443 

Added a new migration in db health check app to remove default permissions for `testmodel`, so they no longer appear in admin panel.


codingjoe (approved):
Hi @aabharya 👋,

Thanks for reaching out!

Sure, why not? I don't see a reason to keep them.

I applied some changes to make sure the migrations are actually run in PyTest.

Cheers!
Joe

PR #452, opened 2025-04-20
Support multiple databases in heartbeat check

RealOrangeOne:
This allows deployments with multiple databases to have the heartbeat run against each database, rather than just the "default" connection.


seanpar203:
@RealOrangeOne A feature for this has been merged into master and released in version 3.19.2

RealOrangeOne:
@seanpar203 that adds it to the main database check, this adds it to the heartbeat check.

PR #469, opened 2025-06-10
Disable atomic requests for health-check view

TheSuperiorStanislav:
Disable atomic requests, because when `ATOMIC_REQUEST=True` django would
still go to db to check the state, meaning there would an error when db is
not available, which would result in django error 500 page rather than
HealthCheck 500 response.

Note: these endpoints are still dependant on db, if you logged in via
browser, or in other words have session cookies.


codingjoe (approved):
Hi @TheSuperiorStanislav 👋,

Thank you for your contribution. I took a look into it, and it didn't work for me just yet. I added a regression test and fixed the implementation.

Furthermore, I will drop the docstring. Documentation is good, but this is rather use case specific, and the behavior is now tested.

Cheers!
Joe

PR #478, opened 2025-07-18
Repair VERSION variable

tacerus:
`VERSION_TUPLE` only resolves to an uninitialized object or tuple (depending on whether type checking is enabled). The real version tuple is contained in `__version_tuple__`.

Fixes: 96dd3bc4aa6e ("Adapt PEP 612 (pyproject.toml) package metadata")


frankwiles:
Awesome thanks for this!

PR #482, opened 2025-08-01
Fix db_heartbeat on Oracle

### LINKED ISSUE #481: health_check.contrib.db_heartbeat does not work on Oracle DB
Oracle doesn't support a `SELECT` without a `FROM` statement in pure SQL. 

To achieve the same behavior as `SELECT 1;` in Oracle, simply use Oracle's native dummy table, called DUAL, i.e., do `SELECT 1 FROM DUAL;`.

Related docs:

https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Selecting-from-the-DUAL-Table.html

Note: On Oracle 23+,  it is now optional to select expressions using the `FROM DUAL `clause
 
https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/Selecting-from-the-DUAL-Table.html

luzfcb:
Oracle doesn't support a `SELECT` without a `FROM` statement in pure SQL. To achieve the same behavior as `SELECT 1;` in Oracle, simply use Oracle's native dummy table, called `DUAL`, i.e., do `SELECT 1 FROM DUAL;`.

https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Selecting-from-the-DUAL-Table.html

Note: On Oracle 23+,  it is now optional to select expressions using the `FROM DUAL `clause
 
https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/Selecting-from-the-DUAL-Table.html

Fix #481


codingjoe (commented):
Hi @luzfcb 👋,

Thanks for reaching out. Oh my, Oracle, my condolences. 😉 I had the great pleasure of implementing returning values in Django, and Oracle almost was the end of me.

Therefore, I am OK if we don't add Oracle to the CI suite to prevent regression.

However, I'd love to change this to an `Expression`, as the implement `as_sql`, `as_oracle`, and more.

It would also allow us to write a test so this never gets deleted.

Cheers!
Joe

PR #492, opened 2025-09-11
fix: Update import for Celery app reference

dnlbauer:
The `default_app` reference used in the Celery Ping healthcheck is `None` for me, leading to this error:

```bash
  File "/usr/local/lib/python3.13/site-packages/health_check/contrib/celery_ping/backends.py", line 15, in check_status
  ping_result = app.control.ping(timeout=timeout)
                        ^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/celery/local.py", line 143, in __getattr__
    return getattr(self._get_current_object(), name)
AttributeError: 'NoneType' object has no attribute 'control'
Internal Server Error: /healthcheck
```

When replacing the import of `default_app` with `current_app`, as shown in the [celery documentation](https://docs.celeryq.dev/en/v5.5.3/userguide/application.html#breaking-the-chain), the health check works for me.

I assume this was a breaking change introduced with Celery 5, but haven't verified this.

Since `current_app` defaults to `default_app` when `default_app` is not `None`, this change should not have a negative impact even for cases where default_app is not None. See  [this line](https://github.com/celery/celery/blob/95bec6d824363a6bb0ff08eda0da42e0d52c2a18/celery/_state.py#L92) in celery.

### DIFF
--- a/health_check/contrib/celery_ping/backends.py
+++ b/health_check/contrib/celery_ping/backends.py
@@ -1,4 +1,4 @@
-from celery.app import default_app as app
+from celery import current_app as app
 from django.conf import settings
 
 from health_check.backends import BaseHealthCheckBackend


insspb:
@dnlbauer Yours celery incorrectly configured. You pointed to documentation where it is pointed, that current app is bad practice.
screen_at_2025-09-19_021111

Read next and you will find out that recommened way will bring you default app in current version of check.

In current_app way you get thread-specific app, that may be different from project-wide. You should check default to ensure that configuration is correct too. Your proposal will break existing checks behaviour in complex systems.

By the way, it is bad check that has a lot of issues in code. Better to check queues only.

codingjoe:
Hi there 👋,

Thanks for reaching out! @dnlbauer is this still an open issue for you, or was @insspb comment sufficient to resolve it?

Please let me know if you need further advice. I will close the PR for now; don't hesitate to reopen it or another issue to further discuss or seek assistance.

Cheers!
Joe

dnlbauer:
> Hi there 👋,
> 
> Thanks for reaching out! @dnlbauer is this still an open issue for you, or was @insspb comment sufficient to resolve it?
> 
> Please let me know if you need further advice. I will close the PR for now; don't hesitate to reopen it or another issue to further discuss or seek assistance.
> 
> Cheers! Joe

Hi,

I didn't found the comment particularly helpful since it didn't provide information in what exactly is wrong.

I implemented my own check using the code from the PR, so it works for me, while the implementation in this library isn't (which very well can be a configuration issue on my side).

codingjoe:
Hm… I see. Sadly, I haven't used Celery in a project myself in a while, and I am not too firm on the issue either.

codingjoe:
Maybe start a discussion in the Q&A section of the repo.

PR #516, opened 2025-12-16
Ref #378 -- Replace deprecated `url` with `path` in documentation

aradkdj:
#378 apparently tried to fix this, but as the docs still show `url` everywhere I'm assuming they are generated from the RST files, not the .md. This PR aims to fix that


codingjoe:
Thanks @aradkdj, much appreciated. Sorry for the delay 👍

PR #545, opened 2026-01-08
Add `HEALTH_CHECK['DISK_USAGE_PATH']` Setting

### LINKED ISSUE #448: Disk Usage Check Fails to Monitor Host Filesystem in Docker Containers
## Description

The current `DiskUsage` health check backend hardcodes the monitored path as [`/`](https://github.com/revsys/django-health-check/blob/9ad0d6586c6e86454be7c7d6fecba6854f07ee6f/health_check/contrib/psutil/backends.py#L19). This works fine when running natively. However, within a Docker container, it checks the container's root filesystem, and not the host's.

This makes the health check less useful in containerized environments, as it doesn't reflect the actual disk usage of the underlying host.

## Suggestion

To enable monitoring the host's disk usage from within a container, a possible approach is to:
  1. Allow configuration of the disk usage path via a setting (e.g., `HEALTH_CHECK['DISK_USAGE_PATH']`).
  2. If `HEALTH_CHECK['DISK_USAGE_PATH']` is set, use that path for `psutil.disk_usage()`. Otherwise, default to `/` for backward compatibility and non-containerized environments.

This would allow users to mount the host's filesystem to a specific path within the container (e.g., `/host/`) and then configure `HEALTH_CHECK['DISK_USAGE_PATH']` to `/host/` to monitor the correct disk usage.

This is the approach used by other monitoring solutions such as Prometheus' [node_exporter](https://github.com/prometheus/node_exporter?tab=readme-ov-file#docker)

---

Happy to send a PR if you agree with the approach.

kilemensi:
Fixes #448.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -75,10 +75,27 @@ HEALTH_CHECK = {
 With the above default settings, warnings will be reported when disk
 utilization exceeds 90% or available memory drops below 100 MB.
 
+By default `psutil` assume the disk to check is available at `/` path.
+To use a different disk e.g. when the app is deployed in a
+container but the relevant disk is the host disk, update the dictionary
+above to include `DISK_USAGE_PATH` setting.
+
+```python
+HEALTH_CHECK = {
+    # existing settings
+    "DISK_USAGE_PATH": "/host/",  # string path to disk
+}
+```
+
 ### `DISK_USAGE_MAX`
 
 Specify the desired disk utilization threshold, in percent. When disk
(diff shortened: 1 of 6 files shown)


codingjoe:
Hi @kilemensi,

Thank you for your contribution. You are right, this was an issue. However, I have been cooking on a larger updated, which has been released in version 3.21.
You can now configure the path and add multiple checks with different pathes, see also:
https://codingjoe.dev/django-health-check/checks/#health_check.Disk

Best
Joe

Resolved in #551

PR #566, opened 2026-01-31
Remove duplicated pyupgrade setup

ulgens:
UP ruleset is already enabled in the ruff config, and ruff is called via hooks. The hook based setup for pyupgrade was redundant.


codingjoe (commented):
Hehehe, I wanted to make sure we stay double up-to-date 😉 
Thanks @ulgens, I appreciate the support <3

codingjoe (approved):

ulgens:
Thanks for the quick response 🌻 Feel free to ping me if any tooling related help is needed.

PR #568, opened 2026-01-31
Add Vary on Accept header

codingjoe:


codingjoe:
@copilot add a regression test.

PR #591, opened 2026-02-03
Adjust management command

TheSuperiorStanislav:
Make `health_check` management command be able to work without making actual HTTP requests.

Before, you could simply run `health_check` command without starting the server, which is more comfortable than keeping in mind that the server must be running. Also, it doesn't work via localhost in k8s pods(Maybe if I specify ip, it would work, but for that I need to specify a different ip for a different project, which is not handy).

So I adjusted the command so that by default it would simulate a request to view(Not calling it via HTTP), but keeping the ability to call it directly via HTTP.

Example of output:

image

### DIFF
--- a/health_check/management/commands/health_check.py
+++ b/health_check/management/commands/health_check.py
@@ -1,19 +1,25 @@
+import http
 import sys
 import urllib.error
 import urllib.request
 
-from django.core.management.base import BaseCommand
-from django.urls import reverse
+from django import test, urls
+from django.core import management
 
 
-class Command(BaseCommand):
+class Command(management.BaseCommand):
     help = "Run health checks and exit 0 if everything went well."
 
-    def add_arguments(self, parser):
+    def add_arguments(self, parser: management.base.CommandParser):
         parser.add_argument(
             "endpoint",
             type=str,
             help="URL-pattern name of health check endpoint to test",
+        )
+        parser.add_argument(
+            "--make-http-request-directly",
+            action="store_true",
+            help="Make a direct http request to server",
         )
         parser.add_argument(
             "addrport",
@@ -23,24 +29,81 @@ class Command(BaseCommand):
             default="localhost:8000",
         )
 
-    def handle(self, *args, **options):
+    def handle(
+        self,
+        *args,
+        **options,
+    ) -> None:
         endpoint = options.get("endpoint")
-        path = reverse(endpoint)
-        host, sep, port = options.get("addrport").partition(":")
+        if options.get("make_http_request_directly"):
+            self.call_endpoint_via_http(endpoint, options.get("addrport"))
+        else:
+            self.call_endpoint_directly(endpoint)
+
+    def call_endpoint_via_http(
+        self,
+        endpoint: str,
+        addrport: str,
+    ) -> None:
+        path = urls.reverse(endpoint)
+        host, sep, port = addrport.partition(":")
         url = f"http://{host}:{port}{path}" if sep else f"http://{host}{path}"
         request = urllib.request.Request(  # noqa: S310
-            url, headers={"Accept": "text/plain"}
+            url,
+            headers={
+                "Accept": "text/plain",
+            },
         )
         try:
             response = urllib.request.urlopen(request)  # noqa: S310
-        except urllib.error.HTTPError as e:
-            # 500 status codes will raise HTTPError
-            self.stdout.write(e.read().decode("utf-8"))
+        except urllib.error.HTTPError as error:
+            self.stdout.write(
+                self.style.ERROR(
+                    error.read().decode("utf-8"),
+                ),
+            )
             sys.exit(1)
-        except urllib.error.URLError as e:
+        except urllib.error.URLError as error:
             self.stderr.write(
-                f'"{url}" is not reachable: {e.reason}\nPlease check your ALLOWED_HOSTS setting.'
+                self.style.ERROR(
+                    f'"{url}" is not reachable: {error.reason}'
+                    "\nPlease check your ALLOWED_HOSTS setting.",
+                ),
             )
             sys.exit(2)
         else:
-            self.stdout.write(response.read().decode("utf-8"))
+            self.stdout.write(
+                self.style.SUCCESS(
+                    response.read().decode("utf-8"),
+                ),
+            )
+
+    def call_endpoint_directly(
+        self,
+        endpoint: str,
+    ) -> None:
+        path = urls.reverse(endpoint)
+        resolved = urls.resolve(path)
+        response = resolved.func(
+            test.RequestFactory().get(
+                path,
+                headers={
+                    "Accept": "text/plain",
+                },
+            ),
+            *resolved.args,
+            **resolved.kwargs,
+        )
+        if response.status_code == http.HTTPStatus.OK:
+            self.stdout.write(
+                self.style.SUCCESS(
+                    response.content.decode("utf-8"),
+                ),
+            )
+        else:
+            for check_result in response.content.decode("utf-8").splitlines():
+                if check_result.endswith("OK"):
+                    self.stdout.write(self.style.SUCCESS(check_result))
+                else:
+                    self.stdout.write(self.style.ERROR(check_result))
+            sys.exit(1)
--- a/health_check/management/commands/health_check.py
+++ b/health_check/management/commands/health_check.py
@@ -3,7 +3,9 @@ import sys
 import urllib.error
 import urllib.request
 
-from django import test, urls
+import asgiref.sync
+from django import http as django_http
+from django import urls
 from django.core import management
 
 
@@ -17,9 +19,9 @@ class Command(management.BaseCommand):
             help="URL-pattern name of health check endpoint to test",
         )
         parser.add_argument(
-            "--make-http-request-directly",
+            "--no-http",
             action="store_true",
-            help="Make a direct http request to server",
+            help="Skip the HTTP stack and perform the checks directly",
         )
         parser.add_argument(
             "addrport",
@@ -35,10 +37,10 @@ class Command(management.BaseCommand):
         **options,
     ) -> None:
         endpoint = options.get("endpoint")
-        if options.get("make_http_request_directly"):
-            self.call_endpoint_via_http(endpoint, options.get("addrport"))
-        else:
+        if options.get("no_http"):
             self.call_endpoint_directly(endpoint)
+        else:
+            self.call_endpoint_via_http(endpoint, options.get("addrport"))
 
     def call_endpoint_via_http(
         self,
@@ -53,6 +55,9 @@ class Command(management.BaseCommand):
             headers={
                 "Accept": "text/plain",
             },
+        )
+        self.stdout.write(
+            self.style.SUCCESS(f"Checking health endpoint at {url}"),
         )
         try:
             response = urllib.request.urlopen(request)  # noqa: S310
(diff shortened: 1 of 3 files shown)


health_check/management/commands/health_check.py
```diff
@@ -1,20 +1,26 @@
...
+    def add_arguments(self, parser: management.base.CommandParser):
         parser.add_argument(
             "endpoint",
             type=str,
             help="URL-pattern name of health check endpoint to test",
         )
+        parser.add_argument(
+            "--make-http-request-directly",
+            action="store_true",
+            help="Make a direct http request to server",
```
codingjoe:
```suggestion
            "--no-http",
            action="store_true",
            help="Skip the HTTP stack and perform the checks directly.",
```

health_check/management/commands/health_check.py
```diff
@@ -23,24 +29,81 @@ def add_arguments(self, parser):
...
+            )
+
+    def call_endpoint_directly(
+        self,
+        endpoint: str,
+    ) -> None:
+        path = urls.reverse(endpoint)
+        resolved = urls.resolve(path)
+        response = resolved.func(
+            test.RequestFactory().get(
```
codingjoe: I don't particularly like using the test framework in production code. `as_view` stores the `view_initkwargs` or the view function. So if you get the view, you can get the checks with `view_fn.view_initkwargs.checks` and run them.

codingjoe (commented):
Hi @TheSuperiorStanislav,

Thank you for trying out the unreleased version and providing invaluable feedback. I sincerely appreciate it, since it helps to ship a sold solution. I was planning on a release candidate this week to invite more people.

That being said, consider opening a discussion before contributing code. It helps to shorten the review cycle and make contributing more fun for you.

Anyhow, I'd generally discourage running the checks without the server running. After all, without the server running, your application isn't healthy. This is particularly concerning, since your WSGI or ASGI server may fill its backlog to the point where it might no longer accept new TCP connections. Load balancers will skip the server, but you might want to recycle it.

That being said, I am not super familiar with k8s, yet I am fairly certain pods provide a loopback interface. I'd appreciate it if you could share a bit more details about the issue you ran into, maybe share some code or traces to help me better understand your motivation and what solution might be available to us.

I am currently hesitant to accept new features, since v4 is supposed to reduce bloat. But I also want to make sure we have solid k8s support 💯 

Cheers!
Joe

TheSuperiorStanislav:
> Hi @TheSuperiorStanislav,
> 
> Thank you for trying out the unreleased version and providing invaluable feedback. I sincerely appreciate it, since it helps to ship a sold solution. I was planning on a release candidate this week to invite more people.
> 
> That being said, consider opening a discussion before contributing code. It helps to shorten the review cycle and make contributing more fun for you.
> 
> Anyhow, I'd generally discourage running the checks without the server running. After all, without the server running, your application isn't healthy. This is particularly concerning, since your WSGI or ASGI server may fill its backlog to the point where it might no longer accept new TCP connections. Load balancers will skip the server, but you might want to recycle it.
> 
> That being said, I am not super familiar with k8s, yet I am fairly certain pods provide a loopback interface. I'd appreciate it if you could share a bit more details about the issue you ran into, maybe share some code or traces to help me better understand your motivation and what solution might be available to us.
> 
> I am currently hesitant to accept new features, since v4 is supposed to reduce bloat. But I also want to make sure we have solid k8s support 💯
> 
> Cheers! Joe

About k8s, the only way I managed to make it work is this

```shell
python manage.py health_check "healthz:health_check_home" `actual-url-to-server`
```

But I also encountered another issue, I'm getting 404 (Unless I change url to https or add an extra `/` to url), and after I fix it am getting this response `Not Acceptable: Supported content types: text/html, application/json`

codingjoe:
> About k8s, the only way I managed to make it work is this [...]

What was the error? As mentioned before, it's difficult to give good advice for a solution without knowing the error.

TheSuperiorStanislav:
> ```shell
> python manage.py health_check "healthz:health_check_home"
> ```

image

codingjoe:
> image

And do you have `localhost` in your `ALLOWED_HOSTS`?

TheSuperiorStanislav:
> > image
> 
> And do you have `localhost` in your `ALLOWED_HOSTS`?

Yes

RahmeVictor:
I want to say that forcing health check to do http requests creates all kinds of problems. You should not have to allow localhost on production to get health checking.

It's even worse if you want to force SSL on all incoming requests.

codingjoe:
> I want to say that forcing health check to do http requests creates all kinds of problems. You should not have to allow localhost on production to get health checking.
> 
> It's even worse if you want to force SSL on all incoming requests.

Hi @RahmeVictor,

Thank you for sharing the feedback. Let me try to explain my reasoning:

Using curl or wget for health checks on containers that run a web server is very common. In fact, so common, [it's literally the example in the official Docker documentation](https://docs.docker.com/reference/dockerfile/#healthcheck), and k8s has a dedicated HTTP probe.

And for good reason, a healthy HTTP server should serve HTTP requests, right?

Therefore, I want to put a determined effort into finding a solution that works for most.
And yes, there will be some scenarios where this won't work. However, they should be so sparse that official support might not be needed, and custom integrations with extensive documentation might be more helpful for the community. So please have some patience with me while we find the best solution for the new (still unreleased) version.


Now, back to k8n: I would recommend trying to use an [HTTP liveness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request). It will call the endpoint for you; no management command required. Make sure to set the correct headers for [X-Forwarded-Host](https://docs.djangoproject.com/en/6.0/ref/settings/#std-setting-USE_X_FORWARDED_HOST) and [SECURE_PROXY_SSL_HEADER](https://docs.djangoproject.com/en/6.0/ref/settings/#secure-proxy-ssl-header) in your probe.

I will try to create a fully working example tomorrow and update the documentation, but no promises. I have a newborn at home, and my FOSS time is limited.

Cheers!
Joe

codingjoe:
Regarding `localost`, yes. That isn't ideal. I was thinking of setting proxy headers; that could be a neat approach for Docker.

RahmeVictor:
> Using curl or wget for health checks on containers that run a web server is very common. In fact, so common, [it's literally the example in the official Docker documentation](https://docs.docker.com/reference/dockerfile/#healthcheck), and k8s has a dedicated HTTP probe.


I agree with the general idea, however there are cases like DHI (Docker hardened images) that don't even ship with curl or wget (or even a shell). It feels a bit like a hack to have to do all of this.

codingjoe:
> I agree with the general idea, however there are cases like DHI (Docker hardened images) that don't even ship with curl or wget (or even a shell). It feels a bit like a hack to have to do all of this.

Hello again,

For that very reason, it is explicitly mentioned in the documentation: https://github.com/codingjoe/django-health-check/blob/b01cffaff8a43a8dd2700110d36a813d33a2f319/docs/container.md?plain=1#L6-L7

If you simply want to check your available memory in k8s and don't care about your HTTP server, please consider [K8s' memory management](https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/).

Best,
Joe

codingjoe:
@TheSuperiorStanislav I updated the verbosity of the management command in 4.0rc3. Can you try this again with `-v2`? I also updated the documentation for K8s to hint towards the HTTP liveliness check.

MatheusGeiger:
This change to the health check command to make one http breaking the behavior of my consumers (i'm using stomp) at deployment kubernetes.

On my scenario I have a command with a consumer stomp and I don't have the web at the same deployment that my consumer do the job. 

Considering scalability, testing, and functionality, it needs to be "alive," connected, and functional on its own. Therefore, it would be beneficial for the test to come from the same deployment where the consumer is running, without depending on an HTTP server.

The old approach in which it was run in bash worked fine when a looking for the long running command process, currently with the http approach it is not possible to run health with the minimal application setup.

receipt example (old model):

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: raizen-chat-messages-consumer-deployment
  namespace: production
  labels:
    app: raizen
    project: loyalty
    bu: varejo
spec:
  replicas: 2
  selector:
    matchLabels:
      app: raizen-chat-messages
  template:
    metadata:
      labels:
        app: raizen-chat-messages
        project: loyalty
        bu: varejo
    spec:
      containers:
        - name: raizen
          image: MY CUSTOM IMAGE APPLICATION
          command: ["./scripts/consumers/start-chat-inbox-consumer.sh"]
          resources:
            limits:
              memory: 868Mi
              cpu: 300m
            requests:
              cpu: 300m
              memory: 356Mi
          envFrom:
            - configMapRef:
                name: raizen-configmap
            - secretRef:
                name: raizen-secret
          env:
            - name: OTEL_SERVICE_NAME
              value: "raizen-chat-messages-deployment"
          startupProbe:
            exec:
              command:
                - /bin/sh
                - -c
                - python manage.py health_check -s readiness
            initialDelaySeconds: 30
            periodSeconds: 45
            timeoutSeconds: 35 # It CPU requests is low, it may take a while to execute, that's why it's 35s
            failureThreshold: 5
          readinessProbe:
            exec:
              command:
                - /bin/sh
                - -c
                - python manage.py health_check -s readiness
            periodSeconds: 45
            timeoutSeconds: 35
```

The "--no-http" seems to be a good alternative for this scenario when you core functionality (workers) doesn't have a web server running, but you still need to run a healthcheck.

codingjoe:
Hi @MatheusGeiger 👋,

Interesting… Thanks for sharing! You are running a Django application that doesn't serve HTTPS but STOMP? Or are those consumers/workers? Out of curiosity, which checks are you running on the container?

Cheers!
Joe

codingjoe:
I created a very minimal patch in 3e66ff8 which has been released in 4.0rc5. However, for now this will remain undocumented. The more information you can give me the more we can improve your documentation to give the right hints for the right use cases.

Thanks for all your contributions! Cheers! Joe

MatheusGeiger:
> Hi @MatheusGeiger 👋,
> 
> Interesting… Thanks for sharing! You are running a Django application that doesn't serve HTTPS but STOMP? Or are those consumers/workers? Out of curiosity, which checks are you running on the container?
> 
> Cheers! Joe

The example is one consumer/worker. This consumers are made by https://github.com/juntossomosmais/django-outbox-pattern library. 

This library are build for us to make more easy create one consumer and publisher using django application to follow the `outbox-pattern`. This works can be "live" and run without any django https server.

Basically i'm have one `long running process` using django management, and this process is the consumer. Internally the consumer uses the STOMP protocol to communicate with the rabbitmq broker. To run the consumer the django http server is not required.

In the application i'm have one custom healthcheck for check the connection for my consumer with the broker. These checks are made to tentative to garantee that my consumer have the correct credentials, host and can connect with my broker.

Looking for production scenario i'm have so many consumers in different kubernetes receipts, to guarantee the readiness for my consumer the connectivity check are made. 

```python
class DjangoOutboxPatternHealthCheck(BaseHealthCheckBackend):
    critical_service = True

    def check_status(self):
        _logger.debug("Checking STOMP server health")
        connection = django_outbox_factories.factory_connection()
        client_id = str(uuid.uuid4())
        # We are using the raw connection given the library is not well-designed for our use case
        # If you come across this code, please check the library's version and see if it has been fixed
        connection_configuration = {
            "username": django_outbox_factories.USERNAME,
            "passcode": django_outbox_factories.PASSCODE,
            "wait": True,
            "headers": {"client-id": f"{client_id}-healthcheck-publisher"},
        }
        try:
            connection.transport.listeners = {}  # Remove all listeners
            connection.connect(**connection_configuration)
            if not connection.transport.is_connected():
                self.add_error(ServiceUnavailable("Unable to connect to STOMP server"))
        except ConnectionRefusedError as e:
            self.add_error(ServiceUnavailable("Connection has been refused"), e)
        except IOError as e:
            self.add_error(ServiceUnavailable("IOError"), e)
        except Exception as e:
            self.add_error(ServiceUnavailable("Unknown error"), e)
        else:
            _logger.debug("Connection established. STOMP server is healthy.")
        finally:
            if connection.transport.is_connected():
                disconnect_receipt = str(uuid.uuid4())
                connection.disconnect(receipt=disconnect_receipt)
```

we already had problems on credentials changes that the healthcheck help us to not have o consumer Downtime. 
The healthcheck fail's with new credentials in new deployment and looking the kubernetes life-cyle the old deployment doesn't is killed when the new deployment have a problem in health probes.

codingjoe:
@MatheusGeiger,

Wow, so interesting. I am always fascinated by the different use cases and a stacks for different projects. Thanks for sharing this in such detail. I hope the latest release candidate will work for you. Please don't hesitate to reach out again.

I want to make sure this package provides a solid foundation for people to build products on. 

Best!
Joe

PR #599, opened 2026-02-03
Fix migration command in migration guide

kojoidrissa:
Correct the migration command for health_check.db to db.

I've updated django-health-check from below 3.21 to 3.21+ twice today. Both time I've run into this error:
`CommandError: No installed app with label 'health_check.db'. Did you mean 'db'?`

Using `migrate db zero` worked in both cases.


codingjoe (approved):
Hi @kojoidrissa,

Thanks for the support! 🥇 

Yes, I double-checked. Poor labeling; good thing it's gone soon.

Cheers!
Joe

codingjoe:
Ported to v3 branch in ea6fbff

PR #607, opened 2026-02-06
Fix #606 -- Assign unique cache key to avoid race conditions

### LINKED ISSUE #606: 🐛 The use of f"itworks-{ts}" in Cache introduces a severe race condition
### Bug Description

Version 3.21 changed the value that Cache check checks from just `itworks` to `itworks-{ts}` where the {ts} is the timestamp. It does the check by calling `set()` and then `get()` without taking a lock of any kind.

This is a textbook race condition that does not even consider the fact the there could be tens of web server instances (e.g. in Kubernetes) sharing the same cluster-wide cache.

### Expected Behavior

I expect the Cache check not have race conditions in the case of shared caches between multiple web servers. Or at least, document this in some way (e.g. tell users to set a host-specific cache key).

### Version

3.21 

### Stack trace

```
2026-02-05 08:56:19,771:    ERROR -- unavailable: Cache key djangohealthcheck_test does not match
  File "[REDACTED]/.venv/lib/python3.11/site-packages/health_check/backends.py", line 43, in run_check
    self.check_status()
  File "[REDACTED]/.venv/lib/python3.11/site-packages/health_check/cache/backends.py", line 44, in check_status
    raise ServiceUnavailable(f"Cache key {self.cache_key} does not match")
health_check.exceptions.ServiceUnavailable: unavailable: Cache key djangohealthcheck_test does not match
2026-02-05 08:56:19,775:    ERROR -- unavailable: Cache key djangohealthcheck_test does not match
  File "[REDACTED]/.venv/lib/python3.11/site-packages/health_check/backends.py", line 43, in run_check
    self.check_status()
  File "[REDACTED]/.venv/lib/python3.11/site-packages/health_check/cache/backends.py", line 44, in check_status
    raise ServiceUnavailable(f"Cache key {self.cache_key} does not match")
health_check.exceptions.ServiceUnavailable: unavailable: Cache key djangohealthcheck_test does not match
```

### Steps to Reproduce

_No response_

samiashi:
Fixes https://github.com/codingjoe/django-health-check/issues/606

`Cache.run()` was using a shared key (`djangohealthcheck_test`) with a per-request value (`itworks-{timestamp}`).
In multi-worker / multi-instance deployments, concurrent probes can overwrite each other between `set` and `get`, producing false negatives:

`ServiceUnavailable: Cache key djangohealthcheck_test does not match`

This PR switches the cache check to an isolated runtime key per probe (`<cache_key>:<uuid4>`), which removes cross-request/key collision races.

A configurable `timeout` with default of 5 seconds on probe keys.  
Reason: with per-probe keys, TTL guarantees automatic cleanup if a process crashes or the check exits before explicit delete, preventing temporary health-check keys from accumulating in shared cache backends.

### DIFF
--- a/health_check/checks.py
+++ b/health_check/checks.py
@@ -8,6 +8,7 @@ import pathlib
 import smtplib
 import socket
 import uuid
+import warnings
 
 import dns.asyncresolver
 import psutil
@@ -54,21 +55,48 @@ class Cache(HealthCheck):
 
     Args:
         alias: The cache alias to test against.
-        cache_key: Prefix for the cache key to use for the test.
+        key_prefix: Prefix for the cache key to use for the test.
+        timeout: Time until probe keys expire in the cache backend.
+        cache_key: Deprecated alias for key_prefix.
 
     """
 
     alias: str = "default"
-    cache_key: str = dataclasses.field(default="djangohealthcheck_test", repr=False)
+    key_prefix: str = dataclasses.field(default="djangohealthcheck_test", repr=False)
+    timeout: datetime.timedelta = dataclasses.field(
+        default=datetime.timedelta(seconds=5), repr=False
+    )
+    cache_key: str | None = dataclasses.field(default=None, repr=False)
+
+    def __post_init__(self):
+        """Support the deprecated `cache_key` argument."""
+        if self.cache_key is None:
+            return
+        warnings.warn(
+            "`Cache.cache_key` is deprecated and will be removed in a future release. "
+            "Use `Cache.key_prefix` instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        if (
+            self.key_prefix != "djangohealthcheck_test"
+            and self.key_prefix != self.cache_key
+        ):
+            raise ValueError("Provide either `key_prefix` or `cache_key`, not both.")
+        self.key_prefix = self.cache_key
 
     async def run(self):
         cache = caches[self.alias]
         ts = datetime.datetime.now().timestamp()
         # Use an isolated key per probe run to avoid cross-process write races.
-        cache_key = f"{self.cache_key}:{uuid.uuid4().hex}"
+        cache_key = f"{self.key_prefix}:{uuid.uuid4().hex}"
         cache_value = f"itworks-{ts}"
         try:
-            await cache.aset(cache_key, cache_value, timeout=30)
+            await cache.aset(
+                cache_key,
+                cache_value,
+                timeout=self.timeout.total_seconds(),
+            )
             if not await cache.aget(cache_key) == cache_value:
                 raise ServiceUnavailable(f"Cache key {cache_key} does not match")
         except CacheKeyWarning as e:
--- a/health_check/checks.py
+++ b/health_check/checks.py
@@ -54,7 +54,7 @@ class Cache(HealthCheck):
 
     Args:
         alias: The cache alias to test against.
-        cache_key: The cache key to use for the test.
+        cache_key: Prefix for the cache key to use for the test.
 
     """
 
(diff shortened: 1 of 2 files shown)


health_check/checks.py
```diff
@@ -64,10 +64,13 @@ class Cache(HealthCheck):
...
         cache = caches[self.alias]
         ts = datetime.datetime.now().timestamp()
+        # Use an isolated key per probe run to avoid cross-process write races.
+        cache_key = f"{self.cache_key}:{uuid.uuid4().hex}"
+        cache_value = f"itworks-{ts}"
         try:
-            await cache.aset(self.cache_key, f"itworks-{ts}")
-            if not await cache.aget(self.cache_key) == f"itworks-{ts}":
-                raise ServiceUnavailable(f"Cache key {self.cache_key} does not match")
+            await cache.aset(cache_key, cache_value, timeout=30)
```
codingjoe: That timeout should be configurable via datetime.timedelta on the dataclass. We have other checks that use timeouts. Best to keep it consistent.

health_check/checks.py
```diff
@@ -64,10 +64,13 @@ class Cache(HealthCheck):
     async def run(self):
         cache = caches[self.alias]
         ts = datetime.datetime.now().timestamp()
+        # Use an isolated key per probe run to avoid cross-process write races.
+        cache_key = f"{self.cache_key}:{uuid.uuid4().hex}"
```
codingjoe: I wonder if we shouldn't just use the node's host name. As in `socket.gethostname`. Although UUIDs won't collide in this universe, and assigning a UUID to a host is pretty intentional. But maybe use `uuid.uuid4()!x` to get all fancy with f-strings ;)

health_check/checks.py
codingjoe: Besides, 30 seems high in this scenario, right?

health_check/checks.py
```diff
@@ -54,7 +54,7 @@ class Cache(HealthCheck):
 
     Args:
         alias: The cache alias to test against.
-        cache_key: The cache key to use for the test.
+        cache_key: Prefix for the cache key to use for the test.
```
codingjoe: Hm… we might want to update the naming to `Cache.key_prefix` then and deprecate the old one. I'd love to keep the naming consistent with the implementation.

health_check/checks.py
```diff
@@ -64,10 +64,13 @@ class Cache(HealthCheck):
...
         ts = datetime.datetime.now().timestamp()
+        # Use an isolated key per probe run to avoid cross-process write races.
+        cache_key = f"{self.cache_key}:{uuid.uuid4().hex}"
+        cache_value = f"itworks-{ts}"
         try:
-            await cache.aset(self.cache_key, f"itworks-{ts}")
-            if not await cache.aget(self.cache_key) == f"itworks-{ts}":
-                raise ServiceUnavailable(f"Cache key {self.cache_key} does not match")
+            await cache.aset(cache_key, cache_value, timeout=30)
+            if not await cache.aget(cache_key) == cache_value:
```
codingjoe, line 79: You are currently targeting main (v4rc3). For the v3 backport, we could simply drop the equation test. It wasn't there before 🤷 What do you think?

codingjoe (commented):
Good stuff! Thanks @samiashi for giving me a hand. Much appreciated <3

health_check/checks.py
samiashi: Sounds good! Added the new key_prefix and deprecated the old cache_key

health_check/checks.py
samiashi: You're right, it is a bit high for this
I've updated it to make it configurable and defaulted the expiry to 5s

health_check/checks.py
samiashi: Good point. I kept UUID-based suffixes because `socket.gethostname()` alone would still collide for concurrent checks on the same node (multiple workers can run probes at the same time). We need uniqueness per probe execution, not just per host

health_check/checks.py
samiashi: I've raised a backport [PR](https://github.com/codingjoe/django-health-check/pull/608) for v3
I've also remove the cache_key and deprecation warning from this as v4 is unreleased and can contain breaking changes. What do you think?

codingjoe (approved):
Excellent! I added a tiny amendment.

health_check/checks.py
```diff
@@ -54,20 +54,30 @@ class Cache(HealthCheck):
...
         alias: The cache alias to test against.
-        cache_key: The cache key to use for the test.
+        key_prefix: Prefix for the node specific cache key.
+        timeout: Time until probe keys expire in the cache backend.
 
     """
 
     alias: str = "default"
-    cache_key: str = dataclasses.field(default="djangohealthcheck_test", repr=False)
+    key_prefix: str = dataclasses.field(default="djangohealthcheck_test", repr=False)
```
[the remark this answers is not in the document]
samiashi: v4 (unreleased) intentionally drops cache_key in favor of key_prefix

health_check/checks.py
```diff
@@ -54,20 +54,30 @@ class Cache(HealthCheck):
...
+    timeout: datetime.timedelta = dataclasses.field(
+        default=datetime.timedelta(seconds=5), repr=False
```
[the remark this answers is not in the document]
samiashi: Updated PR description to reflect the change from 30 to 5 seconds default timeout

PR #608, opened 2026-02-06
Fix #606 -- Assign unique cache key to avoid race conditions

### LINKED ISSUE #606: 🐛 The use of f"itworks-{ts}" in Cache introduces a severe race condition

samiashi:
Backport of 1f5aec50c0a8b0c1ec4b8e5487b46e82c7ccc952
Fix #606

### DIFF
--- a/health_check/cache/backends.py
+++ b/health_check/cache/backends.py
@@ -1,5 +1,4 @@
 import dataclasses
-import datetime
 import uuid
 import warnings
 
@@ -37,6 +36,7 @@ class CacheBackend(HealthCheck):
     cache_key: str | None = dataclasses.field(
         default=getattr(settings, "HEALTHCHECK_CACHE_KEY", "djangohealthcheck_test"), repr=False
     )
+
     def __post_init__(self):
         if self.cache_key:
             warnings.warn(
--- a/health_check/cache/backends.py
+++ b/health_check/cache/backends.py
@@ -50,15 +50,13 @@ class CacheBackend(HealthCheck):
 
     def check_status(self):
         cache = caches[self.alias]
-        runtime_cache_key = f"{self.key_prefix}:{uuid.uuid4().hex}"
-        runtime_cache_value = "itworks"
         try:
             cache.set(
-                runtime_cache_key,
-                runtime_cache_value,
+                self.cache_key,
+                "itworks",
             )
-            if cache.get(runtime_cache_key):
-                raise ServiceUnavailable(f"Cache key {runtime_cache_key} does not match")
+            if cache.get(self.cache_key):
+                raise ServiceUnavailable(f"Cache key {self.cache_key!r} does not match")
         except CacheKeyWarning as e:
             self.add_error(ServiceReturnedUnexpectedResult("Cache key warning"), e)
         except ValueError as e:
--- a/health_check/cache/backends.py
+++ b/health_check/cache/backends.py
@@ -1,5 +1,7 @@
 import dataclasses
 import datetime
+import uuid
+import warnings
 
 from django.conf import settings
 from django.core.cache import CacheKeyWarning, caches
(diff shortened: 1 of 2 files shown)


health_check/cache/backends.py
```diff
@@ -31,17 +33,39 @@ class CacheBackend(HealthCheck):
...
-            cache.set(self.cache_key, f"itworks-{ts}")
-            if not cache.get(self.cache_key) == f"itworks-{ts}":
-                raise ServiceUnavailable(f"Cache key {self.cache_key} does not match")
+            cache.set(
+                runtime_cache_key,
+                runtime_cache_value,
+                timeout=self.timeout.total_seconds(),
+            )
+            if cache.get(runtime_cache_key) != runtime_cache_value:
+                raise ServiceUnavailable(f"Cache key {runtime_cache_key} does not match")
```
codingjoe: Let's keep this as close to v3.20 as possible to avoid further regression: https://github.com/codingjoe/django-health-check/blob/3.20.7/health_check/cache/backends.py
```suggestion
    key_prefix: str = dataclasses.field(default="djangohealthcheck_test", repr=False)
    cache_key: str | None = dataclasses.field(
        default=getattr(settings, "HEALTHCHECK_CACHE_KEY", "djangohealthcheck_test"), repr=False
    )

    def __post_init__(self):
        if self.cache_key:
            warnings.warn(
                "`CacheBackend.cache_key` is deprecated and will be removed in next major release. "
                "Use `CacheBackend.key_prefix` instead.",
                DeprecationWarning,
                stacklevel=2,
            )
        else:
            self.cache_key = f"{self.key_prefix}:{uuid.uuid4().hex}"

    def check_status(self):
        cache = caches[self.alias]
        runtime_cache_key = f"{self.key_prefix}:{uuid.uuid4().hex}"
        runtime_cache_value = f"itworks"
        try:
            cache.set(
                runtime_cache_key,
                runtime_cache_value,
            )
            if cache.get(runtime_cache_key):
                raise ServiceUnavailable(f"Cache key {runtime_cache_key} does not match")
```

codingjoe (commented):
Cool! This safes me time, but let's play this safe.

health_check/cache/backends.py
samiashi: Updated!

codingjoe (approved):
I took the liberty to make some small adjustments.

PR #610, opened 2026-02-06
Ref #608 -- Add previous cache result validation

samiashi:
Apologies there was a major bug in my previous PR that caused the validation to fail incorrectly!
https://github.com/codingjoe/django-health-check/pull/608


codingjoe (approved):
Sorry, I removed some bits. I was under the impression, that the string wasn't checked <3.20. But it seems to be the case.

PR #614, opened 2026-02-09
Make psutil an optional dependency

### LINKED ISSUE #613: 🐛 Allow skipping psutils dependency
### Bug Description

I run the django-health-check on version 3.19 and I never had problem with it, but today, after updating the lib >= 3.23 it started requiring psutils and generated me an outaged on my Django + Zappa + Lambda setup.

Could we make it retrocompatible?

### Expected Behavior

I expected to be able to run it without psutils dependency.

### Version

3.23

### Stack trace

_No response_

### Steps to Reproduce

_No response_

vmesel:
Users who only need database, cache, or other checks no longer have to install psutil. Disk and Memory checks now raise ServiceUnavailable with a clear install instruction when psutil is missing.

Closes #613

### DIFF
--- a/health_check/checks.py
+++ b/health_check/checks.py
@@ -40,6 +40,7 @@ except ModuleNotFoundError:
     class RedisError(Exception):
         pass
 
+
 try:
     import psutil
 except ModuleNotFoundError:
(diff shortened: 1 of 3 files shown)


codingjoe:
Hi @vmesel, in light of my [comment](https://github.com/codingjoe/django-health-check/discussions/616#discussioncomment-15754482) I will close the PR for now. However, feel free to continue the discussion. I want to make sure this works for you. If not we will find another solution. Best, Joe

Issue #428, opened 2024-05-08
Allow CeleryPingHealthCheck to check user-specific queues


BoPeng:
`CeleryPingHealthCheck` checks if some pre-defined queues have active workers

https://github.com/revsys/django-health-check/blob/251b1562c1a7b4bb95b46066fbde0bfaac41567b/health_check/contrib/celery_ping/backends.py#L53

Upon checking, we found that the `CELERY_QUEUES` setting is defined but is empty for our production server. Because this setting is now called [`task_queues`](https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-task_queues) and its documentation suggests against specifying this setting and use automatic routing, we do not want to mess with this option for our server. Can `django-health-check` allow a setting for a list of queues to check?

sevdog:
This is related to #374.

BoPeng:
Yes, but I was expecting a separate setting that allows us to bypass celery configuration since `app.conf.CELERY_QUEUES` is empty if auto-routing is used. I mean, the production server should be flexible enough to dispatch to available workers automatically but healthcheck should alert us when some workers and/or queues are unavailable.

codingjoe:
Hi there 👋,

Thank you for reaching out. I am currently very hesitant to add more settings. Especially for contrib features. The default implementation should be pretty sufficient for most scenarios.

If you need a customization, I would encourage you to use inheritance to apply the changes you need in your project.

Don't hesitate to reopen this issue, but I will close it for now.

Thanks,
Joe

PR #668, opened 2026-02-26
Add "custom_task_queue_names" to Celery Ping check

eht16:
This way users can specify a custom set of queue names which should be checked.

As the Celery documentation recommends, most users probably don't use `app.conf.task_queues` as it is rather complicated and probably not what most users need.

To be able to still check existing queues and not only the default one, it would be very convenient to pass a set of queue names to be checked.

I'm aware of #428 but with the new checks structure it is very easy to add.

What do you think?

### DIFF
--- a/health_check/contrib/celery.py
+++ b/health_check/contrib/celery.py
@@ -28,7 +28,9 @@ class Ping(HealthCheck):
 
     CORRECT_PING_RESPONSE: typing.ClassVar[dict[str, str]] = {"ok": "pong"}
     app: celery.Celery = dataclasses.field(default_factory=app_or_default)
-    custom_task_queue_names: set[str] | None = dataclasses.field(default=None, repr=False)
+    custom_task_queue_names: set[str] | None = dataclasses.field(
+        default=None, repr=False
+    )
     timeout: datetime.timedelta = dataclasses.field(
         default=datetime.timedelta(seconds=1), repr=False
     )
--- a/health_check/contrib/celery.py
+++ b/health_check/contrib/celery.py
@@ -18,6 +18,7 @@ class Ping(HealthCheck):
 
     Args:
         app: Celery application instance to use for the health check, defaults to the [default Celery app][celery.app.default_app].
+        custom_task_queue_names: Set of Celery queue names to be checked if set, defaults to [{}].
         timeout: Timeout duration for the ping command.
         limit: Maximum number of workers to wait for before returning. If not
             specified, waits for the full timeout duration. When set, returns
@@ -27,6 +28,7 @@ class Ping(HealthCheck):
 
     CORRECT_PING_RESPONSE: typing.ClassVar[dict[str, str]] = {"ok": "pong"}
     app: celery.Celery = dataclasses.field(default_factory=app_or_default)
+    custom_task_queue_names: set[str] | None = dataclasses.field(default=None, repr=False)
     timeout: datetime.timedelta = dataclasses.field(
         default=datetime.timedelta(seconds=1), repr=False
     )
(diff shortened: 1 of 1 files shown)


codingjoe:
Hello again, 👋 

Same as before, love the effort. Can you elaborate on your use case a little? I don't really understand why you'd provide a separate queue list. If you are to define an expected queue list, `task_queues` seems the logical choice to me.

If you aren't concerned about your queues and have Celery dynamically manage them, the check should pass without problems too.

Please do consider opening a discussion or issue first next time. It'll make contributing much less difficult, as we can have those discussions ahead of hard code work.

Cheers!
Joe

eht16:
Hey,

> Can you elaborate on your use case a little? I don't really understand why you'd provide a separate queue list. If you are to define an expected queue list, `task_queues` seems the logical choice to me.

In my project we have three queues with some workers on each (dynamically scaled).
Though the queues are not defined explicitly by `task_queues` and so the check falls back to the `task_default_queue` setting and sees only the default queue.

This works fine but obviously it checks only the default queue.

It feels wrong to me to migrate the existing, working code to the more complex `task_queues` setting just for the this health check.

So it would be the easiest to pass the queue names to the check and have them checked all at once.
Of course, this makes sense only if the queues are static.

eht16:
Alternatively, I'd be fine with overriding the check as well.

Then it would be nice to seperate the logic for filling `defined_queues` into an own method so it can be overridden (and tested) more easily.

Something like (untested):
```diff
diff --git a/health_check/contrib/celery.py b/health_check/contrib/celery.py
index 1a20385..a2f79db 100644
--- a/health_check/contrib/celery.py
+++ b/health_check/contrib/celery.py
@@ -59,11 +59,8 @@ class Ping(HealthCheck):
             yield worker
 
     def check_active_queues(self, *active_workers):
-        try:
-            defined_queues = {queue.name for queue in self.app.conf.task_queues}
-        except TypeError:
-            # conf.task_queues may be None
-            defined_queues = {self.app.conf.task_default_queue}
+        defined_queues = self._get_defined_queues()
+
         active_queues = {
             queue.get("name")
             for queues in self.app.control.inspect(active_workers)
@@ -74,3 +71,10 @@ class Ping(HealthCheck):
 
         for queue in defined_queues - active_queues:
             raise ServiceUnavailable(f"No worker for Celery task queue {queue}")
+
+    def _get_defined_queues(self):
+        try:
+            return {queue.name for queue in self.app.conf.task_queues}
+        except TypeError:
+            # conf.task_queues may be None
+            return {self.app.conf.task_default_queue}
```

codingjoe:
Improving the architecture for custom implementation through inheritance is always a good idea. I am also not against adding an argument, if the use case seems somewhat common. I just dont understand it good enough yet. Is the check failing for you somehow?

eht16:
Nope, it is not failing at all.

In my setup, there are three Celery queues and a couple of workers.
Each worker is dedicated to a specific queue and tasks are created for a specific queue.
The queues itself are not configured in Celery explicitly, only the tasks are bound to queues using the `app.conf.task_routes` setting (https://docs.celeryq.dev/en/stable/userguide/routing.html#automatic-routing).

I cannot say if this is the most common way to configure Celery and its queues but it's a simpler way than via `app.conf.task_queues`.

When the `Ping` check tries to fill `defined_queues` with this setup, it will set only the default queue as it doesn't know about the dynamically created queues.

When they can be added manually via the changes in this PR, the check is more extensive by not only checking the default queue on the responding workers.

I think there is no easy and cheap way to query all dynamically created queues from Celery itself without asking every single worker, so defining them by the user seems like a good tradeoff to me.

codingjoe:
Hi @eht16,

Thanks for sharing such a detailed response. I don't use Celery myself too often; I'm more of a dramatiq guy. Hm… considering the uncertainty, I'd prefer to go with inheritance for now, unless more people report issues. If you want to refactor the check to simplify your customization. Go for it!

I will close this PR for now, OK?

Cheers!
Joe

PR #669, opened 2026-02-26
Remove unused "hostname" setting from a couple of checks

eht16:
The `hostname` setting in the `psutil` related checks seem to be a left-over or so and were not used.

### DIFF
--- a/health_check/contrib/psutil.py
+++ b/health_check/contrib/psutil.py
@@ -2,7 +2,6 @@ import dataclasses
 import datetime
 import os
 import pathlib
-import socket
 
 import psutil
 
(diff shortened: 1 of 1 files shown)


codingjoe:
Hi @eht16,

Thank you for your contribution. Maybe consider opening a bug report before jumping into action mode next time. In this particular case, I could have saved you the trouble. The host name is included on purpose. If you have a high availability setup, you never know which server your load balancer will choose. To simplify troubleshooting, we include the hostname in the `__repr__` without any purpose beyond that.

This is also the reason this is only available in the psutil checks, since those are host specific.

Don't hesitate to submit further contributions; I sincerely appreciate it. However, I will close the PR for now. Please also leave a response in a discussion or on this PR.

Cheers!
Joe

eht16:
Alright, sounds reasonable.

A comment might help to remember this in a couple of months and make it clearer to others :).

No worries about the efforts, it didn't take too much time.

codingjoe:
@eht16 documentation updates are always welcome. As a maintainer, you get easily blindsided.

PR #700, opened 2026-04-07
Fix #699 -- Gracefully handle None value from active_queues in Celery

### LINKED ISSUE #699: 🐛 Celery ping check crashes when broker restarts
### Bug Description

🐛 Celery ping check crashes when broker restarts because health_check/contrib/celery.py line 69 self.app.control.inspect(active_workers).active_queues() returns None and the situation is not handled.

### Expected Behavior

When .active_queues() returns None, active_queues should be set to empty set. It is an easy fix. I can make PR if needed.

### Version

4.2.1

### Stack trace

```
AttributeError: 'NoneType' object has no attribute 'values'
  File "health_check/base.py", line 81, in get_result
    ) else await asyncio.to_thread(self.run)
  File "asyncio/threads.py", line 25, in to_thread
    return await loop.run_in_executor(None, func_call)
  File "concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "health_check/contrib/celery.py", line 50, in run
    self.check_active_queues(*self.active_workers(ping_result))
  File "health_check/contrib/celery.py", line 71, in check_active_queues
    .values()
```

### Steps to Reproduce

_No response_

saschabuehrle:
Fixes #699

`inspect(...).active_queues()` can return `None` when the broker is restarting. This makes `check_active_queues()` handle a missing inspect payload as an empty queue mapping, so it reports unavailable queues instead of crashing with `AttributeError`.

Added a regression test for the `None` return path.

Greetings, saschabuehrle

### DIFF
--- a/health_check/contrib/celery.py
+++ b/health_check/contrib/celery.py
@@ -64,11 +64,14 @@ class Ping(HealthCheck):
         except TypeError:
             # conf.task_queues may be None
             defined_queues = {self.app.conf.task_default_queue}
+        inspect_result = self.app.control.inspect(active_workers)
+        active_queues_by_worker = inspect_result.active_queues() if inspect_result else None
+        if not active_queues_by_worker:
+            active_queues_by_worker = {}
+
         active_queues = {
             queue.get("name")
-            for queues in self.app.control.inspect(active_workers)
-            .active_queues()
-            .values()
+            for queues in active_queues_by_worker.values()
             for queue in queues
         }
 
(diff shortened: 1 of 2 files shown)


codingjoe (commented):
Hi @saschabuehrle,

Thanks for reaching out and providing a patch.
I think this can either be solved with a tiny `or {}` change, or you catch the attribute error. Both work for me.

Extra kudos for the test!

Cheers!
Joe

codingjoe (approved):
I took the liberty to amend a commit.

PR #716, opened 2026-04-15
Ref #701 -- Add support for a custom executor for synchronous checks

### LINKED ISSUE #701: 🐛 4.x: memory and DB connection growth under frequent health checks (WSGI / Gunicorn sync workers)
## Summary
After upgrading from **django-health-check 3.24.0** to **4.x** (e.g. **4.2.1**), we observed **rapid ECS task memory growth** and a **large, sustained increase in PostgreSQL connection count** correlated with the deployment. Pinning back to **3.24.0**  brought behaviour back in line with expectations.
I am opening this in case the **4.x execution model** (async view + `asyncio.to_thread()` for sync checks, without the **3.x thread-pool `connections.close_all()`** cleanup) is something you want to address upstream.

---
## Environment
- **django-health-check:** 4.2.1 (problematic) vs 3.24.0 (baseline)
- **Django:** 6.x
- **App server:** Gunicorn **sync** workers (WSGI)
- **Infra:** AWS ECS/Fargate; ALB health checks hitting `/healthcheck/` on an interval of the order of **seconds**
- **Checks in use:** at least **Database**, **Cache**, **Storage** / **Mail** (AWS clients)

---
### Task memory (4.x)
ECS **TaskMemoryUtilization** rose roughly **17% → 77%** over about **~1 hour** (~**1.8 GiB/hour**). Steep, roughly linear climb after the service became active.

Image

### Task memory (3.x)
With **3.24.0**, over **~7 hours** utilization moved about **19% → 23%** (~**0.017 GiB/hour** — much smaller slope than 4.x).

Image

### Database connections
**Before** deploying 4.x: connection count roughly **~5–25** (stable band).
**After** deploying 4.x: baseline and peaks **much higher**, with frequent spikes up to roughly **~110–140** connections (approaching our chart max ~145).

Image

---

## Expected behaviour
- Health checks should **not** cause **unbounded** or **steep linear** process memory growth when probed frequently.
- Health checks should **not** cause a **multiplied, sustained** increase in **open DB connections** compared to 3.x under the same probe rate and worker count.

---
### 1. Thread-local DB connections without `connections.close_all()` in the worker thread
**3.24.0** runs sync plugin checks in a **bounded `ThreadPoolExecutor`** and, when threading is enabled, explicitly closes DB connections **after each check** in the pool thread:
- https://github.com/codingjoe/django-health-check/blob/3.24.0/health_check/mixins.py#L59-L85
**4.2.1** uses an **async** view and runs **non-async** `run()` via **`asyncio.to_thread()`**:
- https://github.com/codingjoe/django-health-check/blob/4.2.1/health_check/views.py#L112-L116  
- https://github.com/codingjoe/django-health-check/blob/4.2.1/health_check/base.py#L76-L93  
There is **no** equivalent **`connections.close_all()`** in that path. Django’s **`connections` are thread-local**; sync checks therefore open DB state on **helper threads**. Combined with **`close_if_unusable_or_obsolete()`** inside checks (does not necessarily close a **new** connection when **`CONN_MAX_AGE`** allows reuse), this plausibly matches **many more live server-side sessions** under frequent probes.
### 2. Process memory (boto3 / botocore + threads)
If **Storage** / **Mail** checks run in those same offloaded threads, **boto3/botocore** TLS/pool state tied to **thread-locals** may contribute to **RSS growth** when probes are hot, in addition to any DB-related effects.
### Note on `asyncio.to_thread`
`to_thread` uses the loop’s **default thread pool** (bounded, reused **within** a given loop). The regression may therefore be less “new thread every time” and more **“sync checks + thread-local resources + missing explicit DB teardown vs 3.x”**, especially under **WSGI + `async_to_sync`** where loop/executor lifecycle may differ from a long-lived ASGI process.

---
## Suggestions for upstream 
1. After running a **sync** check in a worker thread, call **`django.db.connections.close_all()`** (or otherwise ensure thread-local connections for that thread are closed), analogous to 3.x’s `mixins.py` `finally` block.
2. Document **Gunicorn sync + frequent probes** as a **high-risk** combo for 4.x until cleanup is aligned with 3.x.
3. Consider a **supported** code path for a **fully synchronous** health view (opt-in) for WSGI deployments.

### Version

4.2.1 (4.x line in general)

### Stack trace

_No response_

## Reproduction notes (rough)
1. Django app behind ALB, **Gunicorn sync** workers, **`CONN_MAX_AGE`** > 0 (e.g. **300**).
2. Enable **`health_check.db`** (and boto3 / **storage** / **mail**).
3. Hit **`/healthcheck/`** on a **short interval** (e.g. 10–15s) across **multiple workers**.
4. Watch **task memory** and **DB connection count** over **1–3 hours**.

codingjoe:


codingjoe:
@copilot address the unresolved review comments.

codingjoe:
pre-commit.ci run

PR #717, opened 2026-04-15
Clean up storage probe files when validation fails

M-Hassan-Raza:
The storage health check writes a temp probe file, validates it, and then deletes it. Before this change, if the write succeeded but validation failed after that, the temp file could be left behind.

That is usually harmless on local disk, but it is not great behavior on object storage backends where failed probes can leave residue behind over time. This change makes the failed validation path do a best effort cleanup before returning the original error.

The normal success path stays the same. The check still verifies that the backend can actually delete the probe file when the check passes, so this does not loosen the contract, it just fixes the broken failure path.

I also added regression coverage for the cases where the saved file is missing after write, where the saved content does not match, and where cleanup itself fails. In the cleanup failure case we still keep the original validation error and only log the cleanup problem instead of masking it.

Tests run:

```console
uv run pytest -q tests/test_checks.py
uv run pytest -q
uvx ruff check health_check/checks.py tests/test_checks.py
```

Results on my side:

```console
uv run pytest -q tests/test_checks.py passed
uv run pytest -q -> 154 passed, 26 skipped
uvx ruff check health_check/checks.py tests/test_checks.py passed
```


codingjoe:
Hi @M-Hassan-Raza,

Thank you for reaching out, and thank you for your contribution. Since you skipped creating a ticket, allow me to ask more questions to understand the root problem better. Would you mind explaining where the check currently fails and leaves a file behind? The only thing I can think of is the content validation, but I might be missing something.

Cheers!
Joe

M-Hassan-Raza:
> Hi @M-Hassan-Raza,
> 
> Thank you for reaching out, and thank you for your contribution. Since you skipped creating a ticket, allow me to ask more questions to understand the root problem better. Would you mind explaining where the check currently fails and leaves a file behind? The only thing I can think of is the content validation, but I might be missing something.
> 
> Cheers! Joe

Hi Joe,

yeah, you were looking at the right branch.

I dont want to oversell it though, this is a narrow post-write cleanup case, not something I’d call a common failure on healthy local storage.

What I meant is just: `save()` has already written the probe file, then the readback validation fails before `check_delete()` runs. So the realistic cases are on that readback side, like `open()` / `read()` / `exists()` failing after the write already happened, rather than the normal success path somehow forgetting to delete.

I’ve dealt with the same kind of cleanup issue in another codebase (ERP related), where uploaded files are created first and then later setup/processing can still fail, so the code keeps track of created file names and deletes them on the exception path. Not the same code here ofc, just the same general idea: once a write already happened, a later failure should still try to clean up.

So agreed, this is a narrow branch. My point with the patch was only to make that branch symmetric, not to claim this is some frequent prod failure.

Cheers,
Hassan

codingjoe (approved):
@M-Hassan-Raza thanks for the explanation. I took the liberty of pushing another commit in an attempt to simplify the solution based on the details you provided. Would you care to give it a thorough look? Thereafter, I'm happy to release this patch ASAP.

Cheers! Joe

M-Hassan-Raza:
> @M-Hassan-Raza thanks for the explanation. I took the liberty of pushing another commit in an attempt to simplify the solution based on the details you provided. Would you care to give it a thorough look? Thereafter, I'm happy to release this patch ASAP.
> 
> Cheers! Joe

Looks great! Thanks for taking the time to trim some noise, please feel free to merge/release whenever you get the time

PR #725, opened 2026-04-17
Resolve #724 -- Add public dataclass field as OpenMetric label

### LINKED ISSUE #724: Add label in openmetrics export
Currently, when requesting `format=openmetrics`, it returns something like this for integrated checks:

```
django_health_check_status{check="Cache(alias='default')"} 1
django_health_check_status{check="Database(alias='default')"} 1
```

It would cool if we could have

```
django_health_check_status{check="Cache", alias="default"} 1
django_health_check_status{check="Database", alias="default"} 1
```

I can try putting up a PR for this. Thx.

RemiDesgrange:
Closes #724 

With openmetrics format, make the "check" label as the name class, and take all fields (that have repr=True) as a metric label.

### DIFF
--- a/health_check/views.py
+++ b/health_check/views.py
@@ -219,7 +219,8 @@ class HealthCheckView(TemplateView):
         return value.replace("\\", "\\\\").replace('"', '\\"').replace("\n", "\\n")
 
     @staticmethod
-    def abnf_dump(o: dict[str, str]):
+    def abnf_dumps(o: dict[str, str]):
+        """Return ABNF OpenMetic labels as a string suitable for use in a metric name."""
         return ",".join(
             f'{key}="{HealthCheckView.abnf_escape(value)}"' for key, value in o.items()
         )
@@ -236,7 +237,7 @@ class HealthCheckView(TemplateView):
         for result in self.results:
             has_errors |= bool(result.error)
             lines.append(
-                f"django_health_check_status{{{self.abnf_dump(result.check.labels)}}} {not result.error:d}"
+                f"django_health_check_status{{{self.abnf_dumps(result.check.labels)}}} {not result.error:d}"
             )
 
         # Add response time metrics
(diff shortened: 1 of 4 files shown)


health_check/views.py
```diff
@@ -217,6 +218,17 @@ def _escape_openmetrics_label_value(self, value):
...
+    def _openmetrics_labels(self, check):
+        """Build OpenMetrics label set from a HealthCheck dataclass instance."""
+        escape = self._escape_openmetrics_label_value
+        parts = [f'check="{escape(check.__class__.__name__)}"']
+        for field in dataclasses.fields(check):
+            if field.repr:
+                parts.append(
+                    f'{field.name}="{escape(str(getattr(check, field.name)))}"'
+                )
+        return "{" + ",".join(parts) + "}"
```
codingjoe, line 230: Please use generators or list comprehension over using `List.append`.

Please also use `json.dumps` to serialize this and not string concatination.

codingjoe (commented):
Thanks, good first version, but let's give this 120%.

You could consider adding a `json_label` function to the `HealthCheck` class. This could also be reused in other machine-readable responses.

Cheers! Joe

codingjoe:
btw, don't worry about Kafka, their package has a bug.

health_check/views.py
RemiDesgrange: Yeah good point, I also remove the "escape" method to use the `json.dump` method, which does the exact same stuff but not manually.

RemiDesgrange:
I have added the `json_label` method. I did not modify the way json response is rendered to keep the scope of the PR to openmetrics.

codingjoe (approved):
Hi @RemiDesgrange, sorry. I forgot that OpenMetrics uses ABNF for labels, not JSON. I took the liberty and added a commit to get this going. Thanks, Joe

PR #726, opened 2026-04-18
Update deprecated redis_url test case

M-Hassan-Raza:
The test app Redis endpoint was still passing redis_url into the check, but the Redis check now expects client_factory or the deprecated client argument. That made /health/redis/ blow up before the check even ran.

This switches the test route to client_factory and adds a regression test that hits the route and makes sure it returns a normal health response.

Tests:
uv run pytest -q tests/contrib/test_redis.py
uv run pytest -q tests/test_management.py tests/test_views.py
uvx ruff check tests/testapp/urls.py tests/contrib/test_redis.py

### DIFF
--- a/tests/contrib/test_redis.py
+++ b/tests/contrib/test_redis.py
@@ -4,6 +4,7 @@ import os
 from unittest import mock
 
 import pytest
+from django.urls import reverse
 
 pytest.importorskip("redis")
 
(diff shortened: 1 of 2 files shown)


codingjoe (commented):
Hello again,

Yes, that makes sense. I don't know about the extra test, though. It doesn't really add anything, does it?

Cheers,
Joe

M-Hassan-Raza:
> Hello again,
> 
> Yes, that makes sense. I don't know about the extra test, though. It doesn't really add anything, does it?
> 
> Cheers, Joe

Should be good now!

codingjoe (approved):
LGTM, thanks again @M-Hassan-Raza

PR #729, opened 2026-05-06
minor bug: convert cache timeout to integer instead of float

Frodothedwarf:
`self.timeout.total_seconds()` returns a float, and [Django's cache framework expects it to be an integer instead of a float.](https://docs.djangoproject.com/en/6.0/ref/settings/#timeout)

I have converted the float to an integer, and altered the tests to reflect that.

### DIFF
--- a/health_check/checks.py
+++ b/health_check/checks.py
@@ -75,7 +75,7 @@ class Cache(HealthCheck):
             await cache.aset(
                 cache_key,
                 cache_value,
-                timeout=self.timeout.total_seconds(),
+                timeout=int(self.timeout.total_seconds()),
             )
             if not await cache.aget(cache_key) == cache_value:
                 raise ServiceUnavailable(f"Cache key {cache_key} does not match")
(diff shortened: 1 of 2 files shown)


codingjoe:
Hi @Frodothedwarf,

Thank you for reaching out and for your contribution. I always appreciate the support. Since you didn't open a bug ticket first, let's take a step back and do triage.

Can you share what error you are experiencing? I am fairly familiar with Django's cache framework and know that the default value gets cast to an int. However, beyond that, Django doesn't specify a type.

That being said, I presume you did experience some kind of warning, maybe from a 3rd party cache backend? I'd be happy to reach out to them or update the Django documentation to provide better guidance.

Cheers!
Joe

Frodothedwarf:
Hi @codingjoe,

I was stumbling into issue where the health check would come up with weird issues when running in Redis Sentinel mode, and I looked into the code in django-health-check, and saw the timeouts were floats.

It might be that it doesn't really matter, and the cache backend handles the further converting to the expected type automatically, as I switched to Dragonfly instead of Redis Sentinel.

However I haven't seen anything use the cache in Django with floats before, so I thought I might as well just share my contribution to the framework.

codingjoe:
I see. @Frodothedwarf thanks for clarifying. I am a less code, meaning less room for error kinda person. Since the typecast adds more complexity, I'd hold off on this change, unless there is further concern for the values to be strings.

But, to not leave you hanging, I opened a proposal to switch Django fully to typed intervals :)

See also: https://github.com/django/new-features/issues/156

PR #731, opened 2026-05-14
Fix Atlassian incident checks

M-Hassan-Raza:
Atlassian Statuspage incidents are returned by summary.json, not components.json.

This makes the check read the endpoint that includes both components and active incidents, so open incidents are not missed.

Also fixes optional-extra CI coverage for this path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,7 @@ jobs:
           - "rabbitmq"
           - "redis"
           - "rss"
+          - "atlassian"
     steps:
       - uses: actions/checkout@v6
       - uses: astral-sh/setup-uv@v7
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ dependencies = [
 [project.optional-dependencies]
 psutil = ["psutil>=7.2.0"]
 celery = ["celery>=5.0.0"]
-kafka = ["confluent-kafka>=2.0.0"]
+kafka = ["confluent-kafka>=2.0.0", "typing-extensions>=4.0.0"]
 rabbitmq = ["aio-pika>=9.0.0"]
 redis = ["redis>=4.2.0"]
 rss = ["httpx>=0.27.0", "feedparser>=6.0.0"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ dependencies = [
 [project.optional-dependencies]
 psutil = ["psutil>=7.2.0"]
 celery = ["celery>=5.0.0"]
-kafka = ["confluent-kafka>=2.0.0", "typing-extensions>=4.0.0"]
+kafka = ["confluent-kafka>=2.0.0"]
 rabbitmq = ["aio-pika>=9.0.0"]
 redis = ["redis>=4.2.0"]
 rss = ["httpx>=0.27.0", "feedparser>=6.0.0"]
--- a/tests/contrib/test_atlassian.py
+++ b/tests/contrib/test_atlassian.py
@@ -23,7 +23,7 @@ from health_check.exceptions import (
 
 
 def _make_response(components, incidents=None):
-    """Build a minimal /api/v2/components.json payload."""
+    """Build a minimal /api/v2/summary.json payload."""
     return {
         "page": {"id": "test"},
         "components": list(components),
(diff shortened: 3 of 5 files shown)


pyproject.toml
```diff
@@ -45,7 +45,7 @@ dependencies = [
 [project.optional-dependencies]
 psutil = ["psutil>=7.2.0"]
 celery = ["celery>=5.0.0"]
-kafka = ["confluent-kafka>=2.0.0"]
+kafka = ["confluent-kafka>=2.0.0", "typing-extensions>=4.0.0"]
```
codingjoe: This was a bug in `confluent-kafka`, and it has been patched here: https://github.com/confluentinc/confluent-kafka-python/pull/2232
```suggestion
kafka = ["confluent-kafka>=2.0.0"]
```

.github/workflows/ci.yml
```diff
@@ -106,6 +106,7 @@ jobs:
           - "rabbitmq"
           - "redis"
           - "rss"
+          - "atlassian"
```
codingjoe, line 109: Nice, I will move this to a different commit.

codingjoe (commented):
Hi @M-Hassan-Raza,

Thank you for your contribution. I am not sure I understand the problem yet. Is something not working for you? If so, what? I am using those checks myself, and I have yet to experience an issue.

Maybe you can elaborate a little more and maybe consider opening a bug ticket next time so we can do triage in advance.

Cheers!
Joe

M-Hassan-Raza:
> Hi @M-Hassan-Raza,
> 
> Thank you for your contribution. I am not sure I understand the problem yet. Is something not working for you? If so, what? I am using those checks myself, and I have yet to experience an issue.
> 
> Maybe you can elaborate a little more and maybe consider opening a bug ticket next time so we can do triage in advance.
> 
> Cheers! Joe

Hi Joe!!

The issue is that the check reads data["incidents"], but it was fetching /api/v2/components.json. That endpoitn only returns page/components, so incidents is always empty.

The Statuspage /api/v2/summary.json endpoint returns both components and incidents, which matches what the check is already trying to read.

I hit this while checking the live Statuspage payloads. For example, Cloudflare summary.json currently includes an active incident, while components.json has no incidents field, so the old code would report OK.

Happy to trim this PR down if you’d rather keep only the endpoint fix and handle the CI matrix separately.

For verification of my claims you can visit these two URLs to see the difference
https://www.githubstatus.com/api/v2/components.json
https://www.githubstatus.com/api/v2/summary.json

pyproject.toml
M-Hassan-Raza: Ah faiR, my bad i didn't investigate the upstream changes, so i got the CI to green status. Good catch!

codingjoe (commented):
Image

We should add an integration test targeting a real endpoint and make the code less defensive. As in: use data[key], not data.get(key, []), to get proper key errors.

Would you care to continue this patch? If you don't have the time, let me know. I want to release a patch asap.

M-Hassan-Raza:
> Image
> We should add an integration test targeting a real endpoint and make the code less defensive. As in: use data[key], not data.get(key, []), to get proper key errors.
> 
> Would you care to continue this patch? If you don't have the time, let me know. I want to release a patch asap.

hahaha not a problem at all, I'll do that shortly got a little bit of free time :D

M-Hassan-Raza:
@codingjoe If we're looking solid let's get this merged? Happy to make any further changes if needed!

tests/contrib/test_atlassian.py
```diff
@@ -335,6 +340,23 @@ async def test_check_status__request_error(self):
...
+        check = GitHub()
+        async with httpx.AsyncClient() as client:
+            response = await client.get(
+                f"{check.base_url}/api/v2/summary.json",
+                headers={"User-Agent": "django-health-check"},
+            )
+        response.raise_for_status()
+        data = response.json()
+        assert isinstance(data["components"], list)
+        assert isinstance(data["incidents"], list)
```
codingjoe: Hm… I was thinking more of testing the actual check against a URL.

We should also add a test for the KeyError. You might actually want to re-raise it with a better exception message.

codingjoe (commented):
I left you a comment about the tests. Please also drop the `pyproject.toml` changes, as previously mentioned.

I need to run some errands. I will revisit this later today.

tests/contrib/test_atlassian.py
```diff
@@ -335,6 +340,13 @@ async def test_check_status__request_error(self):
 class TestGitHub:
     """Test GitHub platform status health check via Atlassian API."""
 
+    @pytest.mark.integration
+    @pytest.mark.asyncio
+    async def test_check_status__live_summary_endpoint(self):
+        """Verify the check can read a real Statuspage summary endpoint."""
+        result = await GitHub().get_result()
```
codingjoe, line 348: You picked the most unstable one 🫠

codingjoe (approved):
I amended a commit updating the exception types and messages.

Thanks!

PR #744, opened 2026-06-23
Add alias label for Redis health checks

### LINKED ISSUE #743: Redis health checks need distinct OpenMetrics labels for multiple clients
## Problem

`health_check.contrib.redis.Redis` can be configured multiple times with different `client_factory` values, which is useful when a Django app has separate Redis clients for cache, channels, Celery, Constance, etc.

However, the Redis check does not expose an `alias` or any other repr-enabled identifying field. The OpenMetrics renderer builds metric labels from `result.check.labels`, and `HealthCheck.labels` only includes dataclass fields with `repr=True`. Because `Redis.client` and `Redis.client_factory` are `repr=False`, multiple Redis checks all emit the same labels:

```text
django_health_check_status{check="Redis"} 1
django_health_check_status{check="Redis"} 1
django_health_check_status{check="Redis"} 1
```

That makes the individual Redis checks indistinguishable in Prometheus and can create duplicate samples for the same metric/label set in a single scrape.

## Example

```python
checks = [
    (
        "health_check.contrib.redis.Redis",
        {"client_factory": lambda: RedisClient.from_url(settings.CHANNELS_REDIS_URL)},
    ),
    (
        "health_check.contrib.redis.Redis",
        {"client_factory": lambda: RedisClient.from_url(settings.CONSTANCE_REDIS_URL)},
    ),
    (
        "health_check.contrib.redis.Redis",
        {"client_factory": lambda: RedisClient.from_url(settings.CELERY_BROKER_URL)},
    ),
]
```

## Expected behavior

The Redis check should support a stable human-readable label, matching the existing pattern used by checks such as `Cache`, `Database`, and `Storage`, which expose an `alias` field.

For example:

```python
(
    "health_check.contrib.redis.Redis",
    {
        "alias": "channels",
        "client_factory": lambda: RedisClient.from_url(settings.CHANNELS_REDIS_URL),
    },
)
```

Would emit distinct OpenMetrics labels:

```text
django_health_check_status{check="Redis",alias="channels"} 1
```

## Proposed fix

Add an optional `alias: str | None = None` field to `Redis`. This should be backwards-compatible because existing callers can omit it, while apps with multiple Redis checks can opt into distinct labels.

pfcodes:
## Summary

- add an optional `alias` field to the Redis health check
- expose the alias through `HealthCheck.labels` so OpenMetrics can distinguish multiple Redis checks
- document `alias` in the Redis check docstring and Redis configuration examples

Closes #743

## Tests

- `uv run --extra redis --group test pytest tests/contrib/test_redis.py -q`
- `uv run --with ruff ruff check health_check/contrib/redis.py tests/contrib/test_redis.py`

### DIFF
--- a/docs/cookbook.md
+++ b/docs/cookbook.md
@@ -192,7 +192,10 @@ application_checks = [
     # Message brokers & caches
     (
         "health_check.contrib.redis.Redis",
-        {"client_factory": lambda: RedisClient.from_url("redis://localhost:6379")},
+        {
+            "alias": "default",
+            "client_factory": lambda: RedisClient.from_url("redis://localhost:6379"),
+        },
     ),
     (
         "health_check.contrib.rabbitmq.RabbitMQ",
@@ -328,7 +331,10 @@ application_checks = [
     "health_check.Storage",
     (
         "health_check.contrib.redis.Redis",
-        {"client_factory": lambda: RedisClient.from_url("redis://localhost:6379")},
+        {
+            "alias": "default",
+            "client_factory": lambda: RedisClient.from_url("redis://localhost:6379"),
+        },
     ),
     (
         "health_check.contrib.rabbitmq.RabbitMQ",
--- a/docs/install.md
+++ b/docs/install.md
@@ -52,6 +52,7 @@ urlpatterns = [
                 (
                     "health_check.contrib.redis.Redis",
                     {
+                        "alias": "default",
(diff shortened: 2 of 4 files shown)


codingjoe:
Please review my comment on the ticket, while pro problem is in fact a bug, it needs to be sovled by extending the `labels` function.

PR #754, opened 2026-07-30
feat: log failed health checks

### LINKED ISSUE #747: Improve logging for failed health checks
While testing my application on Kubernetes, I configured the `/_healthz/` endpoint as the `startupProbe`.

The probe kept failing, but nothing was logged to indicate which health check was responsible.
The only way I could identify the issue was by manually running the checks from the shell:

```python
>>> from health_check.checks import *; import asyncio; res = asyncio.run(DNS().get_result()); print(res)
HealthCheckResult(..., error=ServiceUnavailable("DNS resolution failed..."), ...)
```

- `get_result()` already captures the exception and returns it in a `HealthCheckResult`, but the error is never logged

Would it be possible to log failed health checks (e.g. at warning or error level)?

This would make debugging much easier and allow us to immediately see which checks fails.

Maybe, add a logger here: https://github.com/codingjoe/django-health-check/blob/2d9b82d6a9abb6be4aa3fd9803b8d3987038b8fc/health_check/views.py#L120 something similar to 
https://github.com/codingjoe/django-health-check/blob/2d9b82d6a9abb6be4aa3fd9803b8d3987038b8fc/health_check/management/commands/health_check.py#L145

Thanks 
Daniel

tomatotomata:
Fixes #747

## Summary

- log handled `HealthCheckException` failures through the shared `health_check.base` logger at WARNING level
- include the failing check class and returned health-check error in the log message
- add regression coverage for the warning record

## Verification

- `uv run pytest tests/test_base.py -q` — 8 passed
- `uv tool run pre-commit run --files health_check/base.py tests/test_base.py` — passed

I also ran the full `uv run pytest` suite. It completed with 158 passed and 26 skipped, plus one existing environment-sensitive DNS assertion: this Windows network returns `no answer` for the test's unroutable resolver (`192.0.2.1`), while the test only accepts messages containing `timeout` or `nameserver`. Rerunning that test reproduces the same unrelated result; this PR does not alter DNS behavior.

AI assistance was used to investigate and draft this focused change; the resulting code and tests were reviewed and verified locally.

### DIFF
--- a/health_check/base.py
+++ b/health_check/base.py
@@ -94,6 +94,7 @@ class HealthCheck(abc.ABC):
             ) else await loop.run_in_executor(executor, self.run)
         except HealthCheckException as e:
             error = e
+            logger.warning("Health check %s failed: %s", self.__class__.__name__, error)
         except BaseException:
             logger.exception("Unexpected exception during health check")
             error = HealthCheckException("unknown error")
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -1,11 +1,12 @@
 import asyncio
 import dataclasses
+import logging
 from unittest.mock import MagicMock, patch
 
 import pytest
 
 from health_check.base import HealthCheck, HealthCheckResult
-from health_check.exceptions import HealthCheckException
+from health_check.exceptions import HealthCheckException, ServiceUnavailable
 
 
 class TestHealthCheck:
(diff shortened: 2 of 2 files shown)


codingjoe:
@ahmadalguydi thanks but your account is too new and you haven't otherwise contributed to this issue. We will not facilitate aura farming, therefor I will reject a PR.

PR #756, opened 2026-08-03
Fix #755 -- Recycle stale database connections before probing

### LINKED ISSUE #755: 🐛 `Database` check reports false failures on stale connections
### Bug Description

I started getting an overwhelming number of false positive failures on my database connction of the form:

```
Unavailable: the connection is closed
```

I / Claude chased this down to it being a stale connection *handle*, which is something that Django can self-recover from.

Here's more details (written by Claude):

## Why it happens

1. **`checks.Database.run` is synchronous**, so `HealthCheck.get_result` dispatches it via
   `loop.run_in_executor(executor, self.run)` (`base.py`), and `HealthCheckView.get_executor`
   returns `contextlib.nullcontext(None)` — so `executor is None` and the work lands on the
   **event loop's default thread pool**. Those threads are long-lived and sit entirely outside
   the request/response cycle.

2. **Django's `connections` is `Local(thread_critical=True)`**, which in a thread with no running
   event loop is a plain `threading.local()` (asgiref `local.py`). So each pool thread keeps its
   own `DatabaseWrapper`. Django's `close_old_connections` is wired to the `request_started` /
   `request_finished` signals, which never fire on those threads — so nothing ever cleans up or
   revalidates a connection parked there.

3. **`Database.run` uses `connection.temporary_connection()`**, which is
   `must_close = self.connection is None`. It closes only what *it* opened. Any connection already
   present on that thread is reused **with no liveness check** — `_cursor()` only pings via
   `close_if_health_check_failed()` when `CONN_HEALTH_CHECKS` is enabled *and* `health_check_done`
   is `False`, and `health_check_done` is reset only by `connect()` and
   `close_if_unusable_or_obsolete()`, neither of which runs on those pool threads.

Net effect: if a connection is present on a default-executor thread and has gone stale, the check
reports a hard failure with no attempt to recover, in a place where Django's usual recovery
machinery is structurally unable to reach.

### Expected Behavior

The check should retrieve/recover from stale connection handles the same way a normal Django application would.

### Version

4.4.1

### Stack trace

_No response_

### Steps to Reproduce

(This section was written by Claude).

Against a real Postgres. Do **not** use sqlite — `sqlite3/base.py::is_usable()` unconditionally
returns `True`, which masks the `CONN_HEALTH_CHECKS` path and makes the bug look different than
it is.

```python
import django
from django.conf import settings

settings.configure(
    DEBUG=False,
    DATABASES={"default": {
        "ENGINE": "django.db.backends.postgresql",
        "NAME": "postgres", "USER": "postgres", "PASSWORD": "postgres",
        "HOST": "localhost", "PORT": "5432",
        "CONN_MAX_AGE": 0,
    }},
    INSTALLED_APPS=["django.contrib.contenttypes"],
    USE_TZ=True,
)
django.setup()

from django.db import connections
from health_check import checks

# Simulate what a server-side termination, a pooler timeout, or a resumed VM
# leaves behind: Django still holds a DatabaseWrapper whose socket is dead.
conn = connections["default"]
conn.connect()
conn.connection.close()          # driver-level close, behind Django's back
assert conn.connection is not None

checks.Database().run()
```

**Expected:** the check notices the handle is unusable, reconnects, and reports OK — the database
is fine and any ordinary Django view would have served the request.

**Actual:**

```
health_check.exceptions.ServiceUnavailable: Unavailable: the connection is closed
```

czue:
Proposed fix for https://github.com/codingjoe/django-health-check/issues/755

Basically adds a call to `close_if_unusable_or_obsolete()` so that stale connections get refreshed automatically.

More details from Claude if you want them:

---
Synchronous checks run on the event loop's default executor, whose threads sit outside the request/response cycle where Django recycles connections. `temporary_connection()` only closes a connection it opened itself, so a connection already parked on one of those threads is reused with no liveness check. Once the server drops that connection, the check reports `Unavailable: the connection is closed` for a perfectly healthy database, then self-clears when the next poll lands on a different thread.

Call `close_if_unusable_or_obsolete()` before probing, mirroring what Django does between requests and what `channels.db.DatabaseSyncToAsync` does for the same class of problem. This also makes CONN_MAX_AGE apply on those threads, where previously nothing ever aged a connection out.

Verified against a real Postgres. sqlite cannot demonstrate this, since its `is_usable()` always returns True and `:memory:` ignores `close()`:

  CONN_MAX_AGE=0                             before: FAILED  after: OK
  CONN_MAX_AGE=60, CONN_HEALTH_CHECKS=True   before: FAILED  after: OK
  CONN_MAX_AGE=60, CONN_HEALTH_CHECKS=False  before: FAILED  after: FAILED

The last row is correct behavior, not a gap. With persistent connections and health checks disabled, ordinary request traffic reuses dead connections too, so a red health check accurately reflects what users experience.

### DIFF
--- a/health_check/checks.py
+++ b/health_check/checks.py
@@ -118,6 +118,12 @@ class Database(HealthCheck):
             connection = connections[self.alias]
         except ConnectionDoesNotExist as e:
             raise ServiceUnavailable("Database alias does not exist") from e
+        # Synchronous checks run on the event loop's default executor, whose threads sit
+        # outside the request/response cycle where Django recycles connections. Without
+        # this, temporary_connection() reuses whatever connection the thread already
+        # holds, even one the server has dropped in the meantime, and the check reports
+        # a failure for a database that is perfectly healthy.
+        connection.close_if_unusable_or_obsolete()
         try:
             compiler = connection.ops.compiler("SQLCompiler")(
                 _SelectOne(), connection, None
--- a/tests/test_checks.py
+++ b/tests/test_checks.py
@@ -141,6 +141,33 @@ class TestDatabase:
         result = await check.get_result()
         assert result.error is None
 
+    @pytest.mark.django_db
+    @pytest.mark.asyncio
+    async def test_run_check__stale_connection(self):
+        """Recycle a connection the database has already dropped."""
+        with mock.patch("health_check.checks.connections") as mock_connections:
+            mock_connection = mock_connections.__getitem__.return_value
+            mock_connection.temporary_connection.side_effect = db.Error(
+                "the connection is closed"
+            )
+            mock_connection.ops.compiler.return_value = mock.MagicMock(
+                return_value=mock.MagicMock(compile=lambda x: ("SELECT 1", []))
+            )
+
+            def revive_connection():
+                mock_connection.temporary_connection.side_effect = None
+                mock_connection.temporary_connection.return_value.__enter__.return_value.fetchone.return_value = (
+                    1,
+                )
+
+            mock_connection.close_if_unusable_or_obsolete.side_effect = (
+                revive_connection
+            )
+
+            assert (await Database().get_result()).error is None, (
+                "Expected the check to recycle the stale connection before probing"
+            )
+
 
 class TestDNS:
     """Test the DNS health check."""


codingjoe:
Hi @czue,

Thank you for reaching out. Django's connection handling is an issue we are well aware of and have spent considerable effort to improve. Please note that we are using temporary connections, which cannot be reused; see also: 2d372f1d380b2c7cfd1d7f16ddb8ea95f6446607

My hunch is that the error you are seeing is not a false positive but an actual production issue. I'd encourage you to open a Q&A discussion to seek help debugging your issue before submitting PRs in the future.

Cheers!
Joe

czue:
Hey @codingjoe thanks for reviewing and sorry for the churn. I was just trying to be a good OSS samaritan but I guess was off base on this one.

> My hunch is that the error you are seeing is not a false positive but an actual production issue

I don't *think* this is accurate. I have sentry hooked up to my app and don't ever see any DB-related issues despite consistent usage. I ended up with this ridiculous patch as a workaround while hoping for an upstream fix, which seems to have addressed the noise I had been seeing with the library.

```python
@dataclasses.dataclass
class Database(checks.Database):
    """Database check that tolerates a stale connection.

    Drop the connection and try once more before reporting the service as down. A database that is
    genuinely unavailable still fails the retry.
    """

    def run(self) -> None:
        try:
            super().run()
        except HealthCheckException:
            if self.alias not in connections:
                raise
            connections[self.alias].close()
            super().run()
```

But, noted for future issues.

codingjoe:
@czue no worries, I always assume the best of intentions, and you even have the street cred to prove it.

Let me try to explain my reasoning; maybe I am off. I am known to make errros. ;)

I create a connection and close it. If there's an open connection on that thread, it's not from this package. And in this case, I am not my brother's keeper. We actually switched to this solution, as I tried pointing out with the commit log. If you allow connections to be reused, having `CONN_HEALTH_CHECKS` certainly makes sense. In this case, Django should discard an unusable connection and open a new one.

I am not 100% if I understood your error matrix correctly, but `CONN_HEALTH_CHECKS=True` worked?

Cheers
-Joe

czue:
no worries! let me try and get better info and i'll come back if I think there's anything actionable in this lib. you're probably right and i'm probably confused...

codingjoe:
@czue sure, but don't hesitate to reach out again. I am no AI, but equally happy to help :P

PR #771, opened 2026-09-17
Add HEAD support to HealthCheckView

noamkush:
Recently gunicorn changed the way it handles `HEAD` with body: https://github.com/benoitc/gunicorn/pull/3710/. As you can see, it gives a warning when a HEAD request has body. This adds an explicit `HEAD` implementation that avoids the body (and content negotiation logic), so gunicorn won't complain about the response body.

Disclaimer: I used an LLM to make the code changes.


codingjoe:
Hi @noamkush, 👋

Thanks for reaching out. I fear this behavior isn't specific to this package.
If, indeed, something needs to be addressed, it'd best be addressed in Django.

I will close the PR for now. Please consider opening a new ticket if I misunderstood the problem or you have any further questions.

Best
Joe

